PR #25, opened 2024-09-10
[Substrate.io Migration] Runtime APIs

CrackTheCode016:
NOTE: This page may be unneeded, but is also good to include just to give light to how this works.  It's possible that after some expansion that it will be something useful.

The "Runtime APIs" page was migrated from Substrate.io:  https://docs.substrate.io/reference/runtime-apis/


nhussein11:
heyy just jumping here to help on this :)

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Runtime APIs
+description: Highlights the runtime interfaces that enable communication with outer node services.  These APIs allow for the outer node to communicate with the WebAssembly runtime of a Polkadot SDK-based node.
+---
+
+Nodes consist of outer node services and a runtime. This separation of responsibilities is an important concept for designing Polkadot SDK-based chains and building upgradeable logic.
```
nhussein11, line 6: I think it would be really helpful to have a little picture or diagram of this concept

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,25 @@
...
+
+For more information on details of how the meta protocol works, [refer to the Polkadot SDK docs.](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/polkadot_sdk/index.html#summary){target=_blank}
+
+However, the outer node services and the runtime must communicate with each other to complete many critical operations, including reading and writing data and performing state transitions.
+
+The outer node services communicate with the runtime by calling runtime application programming interfaces to perform specific tasks.
```
nhussein11, line 12:
```suggestion
The outer node services communicate with the runtime by calling runtime application programming interfaces (or runtime APIs) to perform specific tasks.
```

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,25 @@
...
+
+By default, the runtime provides the following traits for outer node services to call. Below is a non-exhaustive list of common runtime APIs:
+
+- [`AccountNonceApi`](https://paritytech.github.io/polkadot-sdk/master/frame_system_rpc_runtime_api/trait.AccountNonceApi.html){target=_blank}
```
nhussein11, line 16: I'd add a small description for each of these items that clarifies what they do. I found something about this here: https://docs.substrate.io/reference/runtime-apis/

dawnkelly09:
@nhussein11 I tried to beef up this page a bit with more details about the API methods, etc. I still need to spell check and things but, can you please take a look and make sure I'm technically making sense with definitions, types, etc. I dug through the code a good bit but I'm not very familiar with Rust so a second set of eyes would be really appreciated! Thank you!

nhussein11:
@dawnkelly09 thank you! I reviewed this page, and there is only one aspect that I'd like to point out:

The parameter `__runtime_api_at_param__ ` is missing in `AuraAPI` and `AccountNonceApi`; however, I think removing that param from all the methods listed would make sense. I looked for the implementation of that param in the polkadot-sdk, and it's never used when calling those functions. 

As far as I know, `__runtime_api_at_param__ `represents the state of the runtime at a specific block, but the compiler ignores it in all instances by using `_` since these methods are functional without needing to differentiate between specific block states. IMHO, this is something super specific that only a few devs would need to know. 

**TL;DR:**
I'd suggest removing the `__runtime_api_at_param__` from each method, or explaining it at the end of the page as a side note :)

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,493 @@
...
+
+            ---
+
+            `extrinsic` ++"<Block as BlockT>::Extrinsic"++ - the extrinsic to apply
+
+        === "Returns"
+
+            `Result` ++"<ApplyExtrinsicResult>"++ - an inclusion outcome which specifies if this extrinsic is included in this block or not
+
+    ??? function "`apply_extrinsic_before_version_6()` - deprecated"
```
nhussein11, line 131: I'd remove deprecated methods
```suggestion
```

CrackTheCode016: Done

CrackTheCode016:
The dropdowns look much better than the table, thank you!

A little note on this type of documentation: I would be a little careful about documenting some of these API methods specifically in the future, as they could be hard to maintain, more could be added, changed ,etc.  

In this case I think it is fine, and looks good, but again the Rust docs cover this stuff in detail anyways and is auto-generated from the codebase itself. 

> The parameter __runtime_api_at_param__  is missing in AuraAPI and AccountNonceApi; however, I think removing that param from all the methods listed would make sense. I looked for the implementation of that param in the polkadot-sdk, and it's never used when calling those functions.


@nhussein11 is correct here, we should look to removing this, as it is part of the macro-madness of the polkadot sdk and not too relevant in this context 😁 we could just include the return type, or simply reference Rust docs instread.

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,389 @@
...
+
+    ??? function "`account_nonce()` - Get current account nonce of given `AccountId`"
+
+        === "Parameters"
+
+            - `account` ++"AccountId"++ - the account to get nonce for
+
+        === "Returns"
+
+            - `Nonce` ++"u64"++ - index of a transaction
```
eshaben, line 54: wouldn't this be the account's nonce?

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,389 @@
...
+            ??? function "`config` ++"struct"++ - a `BenchmarkConfig` struct"
+                - `pallet` ++"Vec<u8>"++ - the encoded name of the pallet to benchmark
+                - `benchmark` ++"Vec<u8>"++ - the encoded name of the benchmark or extrinsic to run
+                - `selected_components` ++"Vec<(BenchmarkParameter, u32)>"++ - the selected component values to use when running the benchmark
+                - `verify` ++"bool"++ - if true, enable an extra benchmark iteration which runs the verification logic for a benchmark
+                - `internal_repeats` ++"u32"++ - number of times to repeat benchmark within the Wasm environment (versus within the client)
+
+        === "Returns"
+
+            - `Result` ++"Vec<BenchmarkBatch>"++ - the results of a single of benchmark
```
eshaben, line 105: What is a BecnhmarkBatch?

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,389 @@
...
+            `Result` ++"tuple"++ - returns a tuple of two vectors named `BenchmarkList` and `StorageInfo`
+
+    ??? function "`dispatch_benchmark()` - dispatch the given benchmark"
+
+        === "Parameters"
+
+            ??? function "`config` ++"struct"++ - a `BenchmarkConfig` struct"
+                - `pallet` ++"Vec<u8>"++ - the encoded name of the pallet to benchmark
+                - `benchmark` ++"Vec<u8>"++ - the encoded name of the benchmark or extrinsic to run
+                - `selected_components` ++"Vec<(BenchmarkParameter, u32)>"++ - the selected component values to use when running the benchmark
```
eshaben, line 99: What's a BenchmarkParameter?

eshaben (commented):
I'm not sure about providing the parameters and returns for these functions, some of these are a bit half baked. I think for the sake of time in honing in on the types and reviewing all of this, maybe we should just point back to the Polkadot SDK docs instead

dawnkelly09:
I switched this up to just give a description of what each API does and let the user know they can select the API name to go to the Rust docs.

develop/parachain-devs/runtime-development/runtime-apis.md
```diff
@@ -0,0 +1,56 @@
...
+- **[`NominationPoolsApi`](https://paritytech.github.io/polkadot-sdk/master/pallet_nomination_pools_runtime_api/trait.NominationPoolsApi.html){target=_blank}** - provides methods to query various balances, pending rewards, and whether a nomination pool or member needs delegate migration
+- **[`OffchainWorkerApi`](https://paritytech.github.io/polkadot-sdk/master/sp_offchain/trait.OffchainWorkerApi.html){target=_blank}** - for tasks completed by off-chain worker
+- **[`SessionKeys`](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/runtime/apis/trait.SessionKeys.html){target=_blank}** - for generating and querying session keys
+- **[`TaggedTransactionQueue`](https://paritytech.github.io/polkadot-sdk/master/sp_transaction_pool/runtime_api/trait.TaggedTransactionQueue.html){target=_blank}** - for interfering with the transaction queue
+- **[`TransactionPaymentApi`](https://paritytech.github.io/polkadot-sdk/master/pallet_transaction_payment_rpc_runtime_api/trait.TransactionPaymentApi.html){target=_blank}** - helps estimate fees based on transaction properties like weight, length, and extrinsic type
+
+
+
```
nhussein11, line 56:
```suggestion
```

nhussein11 (approved):
LGTM!

eshaben (approved):
LGTM! ✨

PR #54, opened 2024-09-25
add an editor config file

eshaben:
This just adds an editor config file that can be used to overwrite default styling preferences implemented by other formatters.


dawnkelly09 (approved):
yay!

PR #55, opened 2024-09-25
add precommit hook to format code snippets

eshaben:
This PR adds a pre-commit hook that formats the snippets in the `.snippets/code/*` directory using Prettier. It also switches any terminal snippets from `.md` files to `.html`


.github/hooks/pre-commit
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Run Prettier on all files in snippets/code
+echo "Running Prettier on the following files:"
+npx prettier --write ".snippets/code/**/*.{js,ts,json,css,html}"
```
nhussein11, line 5:
```suggestion
npx prettier --write ".snippets/code/**/*.{js,ts,json,css,html,rs}"
```

nhussein11 (approved):
LGTM! 🚀 

I was also thinking that we should add linting for Rust code snippets. Although we don’t have any at the moment, we might have them in the future. I don’t think prettier is meant to lint those files, but we can look into that later, not a priority by any means haha

PR #61, opened 2024-09-25
Add husky

eshaben:
This PR adds husky for running pre-commit hooks to format the code in the snippets directory and cleans up the existing snippets (AUTOMATICALLY 🎉 😂 )


.husky/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+echo "Running Prettier on .snippets/code/**/*.{js,json,html}"
+npx prettier --write .snippets/code/**/*.{js,json,html}
```
nhussein11, line 5:
```suggestion
npx prettier --write .snippets/code/**/*.{js,ts,json,html}
```

.husky/pre-commit
```diff
@@ -0,0 +1,8 @@
...
+
+echo "Adding formatted files back to the commit..."
+git add .snippets/code/**/*.{js,json,html}
```
nhussein11, line 8:
```suggestion
git add .snippets/code/**/*.{js,ts,json,html}
```

.snippets/code/tutorials/tooling/zombienet/overview/base-example.json
```diff
@@ -3,8 +3,6 @@
         "timeout": 1000,
         "bootnode": false,
         "provider": "kubernetes",
-        "backchannel": false,
-        ...
-    },
-    ...
```
nhussein11, line 9: I don't think the pre-commit hook should remove the `...` since it has been added to showcase that there is content before and after this portion of the JSON file.

The same applies for the other instances

nhussein11:
Maybe as a workaround if configuring husky is too complicated, could be to set as a standard one of these two approach when we add json snippets:

- Option 1:
```json
{
    // ...
    "relaychain": {
        "default_command": "polkadot",
        "default_image": "polkadot-debug:master",
        // ...
        "nodes": [
            {
                "name": "alice",
                "validator": true,
                "balance": 1000000000000
            },
            {
                "name": "bob",
                "validator": true,
                "balance": 1000000000000
            }
        ]
        // ...
    }
    // ...
}
```

- Option 2:

```json
{
    "...": {},
    "relaychain": {
        "default_command": "polkadot",
        "default_image": "polkadot-debug:master",
        "...": {},
        "nodes": [
            {
                "name": "alice",
                "validator": true,
                "balance": 1000000000000
            },
            {
                "name": "bob",
                "validator": true,
                "balance": 1000000000000
            }
        ]
    }
}
```

.husky/pre-commit
nhussein11: zombienet uses .toml files to define the spawned networks

nhussein11 (approved):
LGTM, this is great! Thank you 🙏

PR #64, opened 2024-09-26
Fix `create_github_issue` function call in check-repositories workflow

0xLucca:
This PR fixes an indentation bug in the script that creates GitHub issues for outdated repositories. 

Previously, the `create_github_issue` function call was outside the for loop, causing only one issue to be created for the last repository. 

The function call has been correctly indented to ensure an issue is created for each outdated repository in the input JSON file.


nhussein11 (approved):
LGTM!

PR #66, opened 2024-09-26
Polkadot Protocol > Protocol Components > Accounts > Proxy Accounts

kapetan3sid:
Migrated https://wiki.polkadot.network/docs/learn-proxies page


nhussein11:
@kapetan3sid  As discussed, I believe we can close this PR since it makes sense that it lives on the wiki :)

PR #99, opened 2024-10-17
[FIX] - Removing duplicated input elements on termynal snippets

nhussein11:
This PR removes duplicate input elements in the Termynal code snippets of the "Install Dependencies for the Polkadot SDK"


eshaben (approved):
LGTM!

0xLucca (changes requested):
The `Add Trusted Nodes` page was broken by this changes
image

nhussein11:
@0xLucca Leading spaces were automatically modified by prettier, shall we turn it off? @eshaben

0xLucca (commented):
I think we should remove it for HTML files maybe

nhussein11:
Mmm, actually, I don't like the idea of removing .html files from Prettier since IMO it's important to have every single piece of code (or html/CSS) standardized and formatted

I think that initially, this add trusted nodes code snippets should have been written using `<pre>` tags for example, to avoid this formatting conflict

eshaben:
Let me take a closer look

0xLucca (approved):
LGTM!

eshaben (approved):
🚀

PR #116, opened 2024-10-23
Develop > Testing > Runtime

nhussein11:
This PR aims to add the runtime page for the testing section of Develop

### DIFF
--- a/develop/blockchains/testing/runtime.md
+++ b/develop/blockchains/testing/runtime.md
@@ -56,8 +56,6 @@ The tests confirm that:
 - **Pallets modify each other's state** - the second test shows how one pallet can trigger changes in another pallet's internal state, confirming proper cross-pallet interactions
 - **State transitions between blocks are seamless** - by simulating block transitions, the tests validate that the runtime responds correctly to changes in the block number
 
-## Conclusion
-
 Testing pallet interactions within the runtime is critical for ensuring the blockchain behaves as expected under real-world conditions. Writing integration tests allows validation of how pallets function together, preventing issues that might arise when the system is fully assembled.
 
 This approach provides a comprehensive view of the runtime's functionality, ensuring the blockchain is stable and reliable.
--- a/develop/blockchains/testing/runtime.md
+++ b/develop/blockchains/testing/runtime.md
@@ -7,19 +7,19 @@ description: Explore runtime testing in Polkadot SDK to verify how multiple pall
 
 ## Introduction
 
-In the Polkadot SDK, it’s important to test individual pallets in isolation and how they interact within the runtime. Once unit tests for specific pallets are complete, the next step is integration testing to verify that multiple pallets work together correctly within the blockchain system. This ensures that the entire runtime functions as expected under real-world conditions.
+In the Polkadot SDK, it's important to test individual pallets in isolation and how they interact within the runtime. Once unit tests for specific pallets are complete, the next step is integration testing to verify that multiple pallets work together correctly within the blockchain system. This testing ensures that the entire runtime functions as expected under real-world conditions.
 
 This article extends the [Testing Setup](/develop/blockchains/testing/setup){target=\_blank} guide by illustrating how to test interactions between different pallets within the same runtime.
 
 ## Testing Pallets Interactions
 
-Once the test environment is ready, tests can be written to simulate interactions between multiple pallets in the runtime. Below is an example of how to test the interaction between two generic pallets, referred to here as `pallet_a` and `pallet_b`. For the coherence of this article, the pallets are assumed to be coupled in the sense that `pallet_b` depends on `pallet_a`. The configuration of `pallet_b` is the following:
+Once the test environment is ready, you can write tests to simulate interactions between multiple pallets in the runtime. Below is an example of how to test the interaction between two generic pallets, referred to here as `pallet_a` and `pallet_b`. In this scenario, assume that `pallet_b` depends on `pallet_a`. The configuration of `pallet_b` is the following:
 
 ```rust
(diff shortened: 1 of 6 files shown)


0xLucca (commented):
The page looks good, but it feels like a tutorial. Maybe a section can be added with more generic examples on how to test storage, events, balances,etc

nhussein11:
> The page looks good, but it feels like a tutorial. Maybe a section can be added with more generic examples on how to test storage, events, balances,etc

well, those elements were covered on the "Pallet Testing" page. Do you believe that they should be included here anyway?

RemyLeBerre:
> > The page looks good, but it feels like a tutorial. Maybe a section can be added with more generic examples on how to test storage, events, balances,etc
> 
> well, those elements were covered on the "Pallet Testing" page. Do you believe that they should be included here anyway?

Perhaps we can just reference it instead? If it's already included in the 'Pallet Testing Page', then simply just referring to it should be fine imho?

"For more details and examples on how to test storage, events, balances etc, please see the 'Pallet Testing Page'."

CrackTheCode016:
Unfortunately, I don't have a lot of experience with this sort of production testing, but the following makes sense to me: 

1. Pallet testing - for events, storage, specific extrinsics etc. 
2. Runtime mocking / integration testing - for testing the cohesion of the system, how it comes together with all pallets, not any one pallet in particular.  This means that a test will test more than one pallet at a time.

It seems like this page reflects that, the Astar example is a very good example of this. As long as we make this distinction clear (which it is for the most part) I think it is fine. Might be a good idea to tag someone on their team to take a look as well, to ensure we've captured this correctly

0xLucca:
> > The page looks good, but it feels like a tutorial. Maybe a section can be added with more generic examples on how to test storage, events, balances,etc
> 
> well, those elements were covered on the "Pallet Testing" page. Do you believe that they should be included here anyway?

You are right. I agree with Bader's comment

nhussein11:
@0xLucca @CrackTheCode016 So, I added a reference to the `pallet testing` page for more explanations about storage, errors, and events testing. I will contact someone from Astar so they can also look at this content to see if it is correct. Meanwhile, I believe we can proceed with formatting reviews to avoid leaving this PR forgotten if you are okay with it

Dinonard:
> @0xLucca @CrackTheCode016 So, I added a reference to the `pallet testing` page for more explanations about storage, errors, and events testing. I will contact someone from Astar so they can also look at this content to see if it is correct. Meanwhile, I believe we can proceed with formatting reviews to avoid leaving this PR forgotten if you are okay with it

Hey @nhussein11, Alberto from Moonbeam asked us to check this PR.
Looks good!

I have a suggestion for the integration test example - perhaps it would be better if you had 2 pallets which aren't tightly coupled, but instead rely on the dependency injection. E.g.:
*  define some trait like `trait Observer { fn notify(n: u32); }`
* `PalletA` has a configurable type that has to implement `trait Observer` and call it each time some action is done (e.g. an extrinsic call)
* `PalletB` provides an implementation of the `trait Observer` which will e.g. store the _notified_ value and/or do something with it (e.g. store the squared value)

The reason why I'm suggesting this is since it's a much more common pattern in FRAME than tight coupling.

nhussein11:
hey @Dinonard! Thank you very much; it makes sense. I jotted this down and will implement it after the first release when we have more bandwidth

Thank you for taking the time to look at this 🙏

dawnkelly09 (commented):
completed grammarly pass and made some formatting edits. Needs final signoff from Erin.

develop/blockchains/testing/runtime.md
```diff
@@ -1,4 +1,63 @@
...
+--8<-- "code/develop/blockchains/testing/runtime/pallet-coupling.rs"
+```
+
+And also, `pallet_b` exposes a call that interacts with `pallet_a`:
+
+```rust
+--8<-- "code/develop/blockchains/testing/runtime/pallet-b-call.rs"
+```
+
+In this first test, a call to `pallet_a` is simulated, and the internal state is checked to ensure it updates correctly. The block number is also checked to ensure it advances as expected.
```
eshaben, line 28:
```suggestion
In this first test, a call to `pallet_a` is simulated, and the internal state is checked to ensure it updates correctly. The block number is also checked to ensure it advances as expected:
```

develop/blockchains/testing/runtime.md
```diff
@@ -1,4 +1,63 @@
...
+
+## Verifying Pallet Interactions
+
+The tests confirm that:
+
+- **Pallets initialize correctly** - at the start of each test, the system should initialize with block number 0, and the pallets should be in their default states
+- **Pallets modify each other's state** - the second test shows how one pallet can trigger changes in another pallet's internal state, confirming proper cross-pallet interactions
+- **State transitions between blocks are seamless** - by simulating block transitions, the tests validate that the runtime responds correctly to changes in the block number
+
+## Conclusion
```
eshaben, line 59:
```suggestion
```

I don't think we need to have a conclusion section here. We can just add the summary at the end of the previous section. Mostly saying this because I feel like if we're going to have a conclusion on one page, we should be consistent and add it to all other pages, but I just don't think that is necessary.

eshaben (commented):
Couple minor things, but LGTM

develop/blockchains/testing/runtime.md
nhussein11: Agree! Fixed, thank you!! 🚀

eshaben (approved):
LGTM! Nice work!

0xLucca (approved):
LGTM!

PR #117, opened 2024-10-23
[Fix] Adding .vscode to .gitignore

nhussein11:
This PR adds `.vscode` folder to `.gitignore` to avoid accidentally committing any local config that any contributor might have in their local dev environment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.vscode


CrackTheCode016 (approved):

eshaben (approved):
LGTM!

PR #166, opened 2024-11-14
update links and formatting on asset conversion page

eshaben:
I came across this page because there were some broken links on it then I went down the rabbit hole and started formatting some stuff on it 😅 

Not sure if we still need this page or what, as it is not in the new IA. But since I had already made the changes before realizing that, I figured I'd create a PR and then we can discuss whether this is needed or not


nhussein11 (approved):
LGTM!

dawnkelly09 (approved):
lgtm

0xLucca (approved):
LGTM!

PR #169, opened 2024-11-14
updates folders and pages to match current IA

dawnkelly09:
This PR should now reflect the correct folders and pages to match the current IA.


develop/blockchains/custom-blockchains/.pages
```diff
@@ -1,8 +1,9 @@
 title: Custom Blockchains
 nav:
   - index.md
-  - 'Overview': overview.md
+  - 'Overview (FRAME)': frame-overview.md
   - 'Add Existing Pallets': add-existing-pallets.md
   - 'Make a Custom Pallet': make-custom-pallet.md
+  - 'Add Smart Contract Functionality': add-smart-contract.md
```
eshaben, line 7:
```suggestion
```

Delete for now

develop/blockchains/custom-blockchains/add-smart-contract.md
```diff
@@ -0,0 +1,6 @@
+---
```
eshaben, line 1: Delete

eshaben (commented):
Awesome! Thanks for tackling this and sorting it all out. Left comments on the empty files we should get rid of now

eshaben:
I went through the IA and deleted any pages where we don't have content. We can add these as we go, but I don't want to have to go back and delete pages again at the end of the week. I left a select few empty pages because we NEED to get those done

nhussein11 (approved):
LGTM!

PR #179, opened 2024-11-19
remove icons from integration pages

eshaben:
This PR removes the placeholder icons from the integration pages. We can worry about adding the brand logos instead after launch


nhussein11 (approved):
LGTM!

dawnkelly09 (approved):
lgtm

PR #183, opened 2024-11-21
Structural changes to root level pages

eshaben:
This PR includes some structural changes to the root-level sections with one page, and switches them to just be root-level pages now that the `polkadot-mkdocs` repo is being updated to support it. The support for it is in [PR 32](https://github.com/papermoonio/polkadot-mkdocs/pull/32), so please when you review this PR, make sure that you pull down the changes from the `eshaben/improvements` branch of the `polkadot-mkdocs` repo.

This also includes a script to add a specific class to root-level index pages, which allows the styling of the navigation menu to be more easily changed. There is a good chance we'll get rid of this in the future, but for now it does the job


albertov19 (changes requested):
LGTM!

One small detail. For example, Development Pathways being a single page, I would like to see the entire navigation path on the left hand side is shown. For example (done in Paint):

image

albertov19 (approved):
LGTM! Sweeet!

dawnkelly09 (approved):
🚀

PR #200, opened 2024-11-22
Set up the bare minimum index pages

eshaben:
This autogenerates the index pages and updates the template to be `index-page.html`, as this is the new template in the mkdocs repo.

This should be reviewed alongside this mkdocs PR: https://github.com/papermoonio/polkadot-mkdocs/pull/34

Note: there will be additional PRs to follow that customize the index pages more.


albertov19 (changes requested):
I did notice that there an content index being shown in the right hand side of the index pages.

image


Unsure if this is a fix here or in the mkdocs repo

albertov19 (approved):

dawnkelly09 (approved):
🚀

PR #208, opened 2024-11-24
Minor fixes

eshaben:
This PR fixes the following:

- Scrolling behavior on Safari. It was pretty jumpy, these changes smooth it out without impacting the scrolling behavior on other browsers
- Removes the index page content for the home page. The home page content is in the `home.html` template in the mkdocs repo, so this is unnecessary and gets overwritten (was added by accident). However, it was also causing issues with the left navigation on smaller screens, next to the home menu option it was showing a TOC icon

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,11 +8,3 @@ hide:
 - navigation
 template: home.html
 ---
-
-# Polkadot Developer Docs
-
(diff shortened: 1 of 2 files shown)


nhussein11 (approved):
LGTM! 🚀

0xLucca (approved):
LGTM!

PR #214, opened 2024-11-24
Delete duplicate and no longer relevant content and images

eshaben:
This PR removes some old stuff that has either been moved to a new home already, or is no longer relevant.

Some of this, if not all of it, was already deleted and somehow reintroduced again. So we need to be careful with handling merge conflicts and making sure we're keeping and deleting the correct files.


0xLucca (approved):
LGTM!

nhussein11 (approved):
LGTM!

PR #256, opened 2024-12-13
[FIX] - Adding `npm install` to contributing.md

nhussein11:
This PR aims to include the command `npm install` in the set of instructions for contributing to polkadot-docs


.CONTRIBUTING.md
```diff
@@ -66,7 +66,13 @@ To set up the structure, follow these steps:
...
+2. Switch to the polkadot-docs directory and use [npm](https://docs.npmjs.com/about-npm) to install dependencies:
+
+    ```bash
+    cd polkadot-docs && npm install
+    ```
+
+3. In the `polkadot-mkdocs` folder, you can build the site by running:
 
     ```bash
     mkdocs serve
```
dawnkelly09, line 78:
```suggestion
    cd ..
    mkdocs serve
```

dawnkelly09: Let's get them back to the right folder safely

dawnkelly09 (changes requested):
one small thing!

.CONTRIBUTING.md
nhussein11: Yeah, I considered adding `cd ..`, but I'm a bit unsure about it.

I was worried about misleading users into thinking they can serve the site only if they run `cd ..` first (which might seem silly, but if it's all together, it could lead to that). WDYT?

0xLucca (approved):
LGTM!

.CONTRIBUTING.md
dawnkelly09: I can see that. Maybe something like "Navigate back to the `polkadot-mkdocs` folder then build the site...."?

nhussein11:
@dawnkelly09 I merged master here and fix the last step, should be good to go finally 🚀

dawnkelly09 (approved):
🚀

PR #259, opened 2024-12-17
add script to fix created dates

eshaben:
This just adds a script to default to the last updated date instead of the current date if for whatever reason the created date cannot be found

Goes with PR: https://github.com/papermoonio/polkadot-mkdocs/pull/56


nhussein11 (approved):
LGTM!

0xLucca (approved):
LGTM!

PR #283, opened 2024-12-24
Minor fix: Moved to inline from being copyable

Jan-Jan:
Problem: Websocket address was presented like it should be copied (which implied it needs to be pasted, which leads down a rabbit hole...). Instead moved it to inline, so that the instruction is clearly to open Polkadot app


CrackTheCode016 (approved):
thank you!

0xLucca (approved):
LGTM!

PR #311, opened 2025-01-20
Update chopsticks docs

muddlebee:
updated the commands

`npx @acala-network/chopsticks --config=polkadot` doesn't work without appending `--config=polkadot` 

and removed the specific version to always use the latest chopsticks version


0xLucca:
Hey @muddlebee! Thank you for your contribution! 

We prefer to keep specific versions in our docs to ensure reproducibility, rather than using "latest". We're actively working on version updates and will revise the docs accordingly.

As for the command issue you reported - I'll investigate that and follow up soon.

0xLucca:
I've created a new PR (#335) that addresses the command issue by showing how to use the `--help` flag to explore available options, while maintaining our versioning approach. I'm closing this PR in favor of that solution.

PR #335, opened 2025-01-28
Add --help flag to command in Chopsticks  page

0xLucca:
Previously, running `npx @acala-network/chopsticks` without flags or commands would hang without providing useful information. Updated the command showing how to run it with the `--help` flag to display available options and commands, making it more helpful for new users.


nhussein11 (approved):

dawnkelly09 (approved):
🚀

muddlebee:
nice. Thank you.

PR #348, opened 2025-01-31
clean up the general management docs

eshaben:
This PR cleans up the following page: https://docs.polkadot.com/infrastructure/running-a-validator/operational-tasks/general-management/

In particular it does the following:
- Makes the list items concise
- Formats each of the lists in the same way
- Updates usage of Alertmanager, so we're not using unnecessary backticks


0xLucca (approved):
LGTM!

dawnkelly09 (approved):
🚀

PR #394, opened 2025-02-17
[FIX] - Pre-commit hook over staged files

nhussein11:
In https://github.com/polkadot-developers/polkadot-docs/pull/390, we've noticed that the pre-commit hook takes too long to execute because the formatters (mainly Taplo) check every file, even if they are already correctly formatted.  

We can leverage `git diff --cached` to apply the pre-commit hook only to files that have changed, significantly reducing the commit time.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,38 +1,53 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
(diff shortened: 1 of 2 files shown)


eshaben (approved):
Thank you!

eshaben (approved):

PR #477, opened 2025-03-18
XCMv5 guide

franciscoaguirre:
Added a section for XCM guides. In there I added a first section called `From apps` and `PAPI` which are meant to show how to use XCM from an app using PAPI. I plan to expand those guides in the future with `From smart contracts` and `From runtimes` also showing different tools. All of these guides will use XCMv5.

Other than that, I created a section on versions where I list differences between versions. I started it with highlighting the differences between v4 and v5.

### DIFF
--- a/develop/interoperability/.pages
+++ b/develop/interoperability/.pages
@@ -3,9 +3,9 @@ nav:
   - index.md
   - 'Introduction to XCM': intro-to-xcm.md
   - guides
-  - 'XCM Channels': xcm-channels.md
-  - 'XCM Configuration': xcm-config.md
+  - 'Channels': xcm-channels.md
+  - 'Runtime Configuration': xcm-config.md
   - 'Send Messages': send-messages.md
-  - 'XCM Runtime APIs': xcm-runtime-apis.md
+  - 'Runtime APIs': xcm-runtime-apis.md
   - 'Test and Debug': test-and-debug.md
   - versions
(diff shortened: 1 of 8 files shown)


nhussein11 (commented):
Reviewed some files and stopped the review here to do not bombard you with too many comments, pls feel free to tag me for a re-review at any time

.snippets/code/develop/interoperability/versions/v5/teleport-and-transact.ts
```diff
@@ -0,0 +1,175 @@
+// `ahp` is the name we gave to `npx papi add`
```
nhussein11, line 1: Pls check the comment about the `we` words and apply it to the snippets as well, just to keep consistency

brunopgalvao (commented):
Excellent content.

Some follow-up content that I think is needed in our Docs:

A full, "holy-grail" teleport example utilizing all the best practices:
- Fee estimation (`TransactionPaymentApi`, XcmRuntimeApi`, etc)
- `AssetClaimer`
- `RefundSurplus`
- Dry run with proper checks e.g. account existence / ED

Teleporting a non-sufficient to Asset Hub
- Using `batch`
- Fee estimation
  - Estimating fees in the non-sufficient token using `quote_price_exact_tokens_for_tokens`
- Using `ExchangeAsset` instruction
  - Swapping for ED and for paying fees

Issue #446, opened 2025-03-12
Add a page or a section on Light Clients


DrW3RK:
https://wiki.polkadot.network/docs/build-light-clients or similar content

PR #479, opened 2025-03-18
Light Client Content / Page

CrackTheCode016:
Closes issue #446 

Adds a new page for covering light clients, namely: 

- What they are, along with how they work
- Common implementations and ways of using light clients in Polkadot


nhussein11:
Hi @CrackTheCode016, thank you very much for this, and sorry for getting on this with such delay! This is great, it's valuable content and looks very interesting to me so far. As soon as we have the bandwidth, we'll merge this into another branch and address all the nitpicky formatting details we look for, so you don't have to worry about that. (I will tag you once it's ready so you can confirm the content it's accurate)

Thanks again!

PR #559, opened 2025-04-11
[FIX] - Rollback json rpc api endpoint

nhussein11:
This PR fixes the endpoint used in https://papermoonio.github.io/polkadot-mkdocs/develop/smart-contracts/json-rpc-apis/, since when executing an eth request to https://westend-asset-hub-rpc.polkadot.io it would not work. Instead, the proper endpoint to use when calling eth methods is - https://westend-asset-hub-eth-rpc.polkadot.io.


eshaben (approved):

dawnkelly09 (approved):
🚀

PR #583, opened 2025-04-28
Update Zero to Hero tutorial to parachain template stable2412

0xLucca:
Update the Zero to Hero tutorial to use [Parachain Template stable2412](https://github.com/paritytech/polkadot-sdk-parachain-template/tree/stable2412)


tutorials/polkadot-sdk/parachains/zero-to-hero/set-up-a-template.md
```diff
@@ -60,13 +60,12 @@ The [Polkadot SDK Parachain Template](https://github.com/paritytech/polkadot-sdk
 
 1. Clone the template repository:
     ```bash
-    git clone https://github.com/paritytech/polkadot-sdk-parachain-template.git parachain-template
+    git clone -b stable2412 https://github.com/paritytech/polkadot-sdk-parachain-template.git parachain-template
```
nhussein11, line 63: What about binding this to a tag instead of a branch? If the branch gets updated for some reason, we take the risk that the snippets might break, right?  


Maybe we can tie this to -> https://github.com/paritytech/polkadot-sdk-parachain-template/releases/tag/stable2412

0xLucca: That command is actually cloning from that specific tag. There is no branch called `stable2412`

nhussein11: Ah, I see, `-b` is for branches, but as long as there are no branches with the same name as the tag, it will clone the tag

This is nitpicking, but maybe we should start enforcing this syntax for these situations: 

```
git clone -b refs/tags/<tag_name> <repository_url>
```

Although it may lack aesthetic appeal, we want to clarify that we are discussing the tag. We can revisit this topic later anyways

eshaben (approved):

PR #589, opened 2025-04-30
Add retry to llms script

0xLucca:
This PR adds the logic to retry the request to pull remote snippets when getting 429 errors

### DIFF
--- a/scripts/generate_llms.py
+++ b/scripts/generate_llms.py
@@ -140,7 +140,7 @@ def fetch_remote_snippet(snippet_ref, yaml_data, max_retries=3, backoff_factor=2
         try:
             response = requests.get(url)
             if response.status_code == 429:
-                wait = backoff_factor ** attempt
+                wait = backoff_factor * attempt
                 print(f"429 Too Many Requests. Retrying in {wait} seconds...")
                 time.sleep(wait)
                 continue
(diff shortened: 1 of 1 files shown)


scripts/generate_llms.py
```diff
@@ -129,30 +130,36 @@ def fetch_remote_snippet(snippet_ref, yaml_data):
...
-
-        return snippet_content.strip()
-    except requests.RequestException as e:
-        print(f"Failed to fetch snippet from {url}: {e}")
-        return f"Error fetching snippet from {url}"
+    for attempt in range(1, max_retries + 1):
+        try:
+            response = requests.get(url)
+            if response.status_code == 429:
+                wait = backoff_factor ** attempt
```
nhussein11, line 143: Isn't a lot to do:
```
 wait = backoff_factor ** attempt
```

Wouldn't: `wait = backoff_factor * attempt` be enough?

0xLucca: Probably exponential backoff is kinda overkill. We can try with linear backoff and see how it works

albertov19:
I'm not against this - is it working? How long it takes?

0xLucca:
It works as expected and takes almost the same time as before. Previously, when a request returned an error, the script just skipped that file, making the `llms.txt` file generation not ok.

albertov19 (approved):
LGTM

eshaben (approved):

PR #664, opened 2025-05-26
[Fix] - Block explorers duplication

nhussein11:
This PR fixes the block explorers duplicated page and also moves contract verification within the new block explorers folder


0xlukem (approved):
LGTM

PR #832, opened 2025-07-21
Add SubWallet

brunopgalvao:
Merging from https://github.com/polkadot-developers/polkadot-docs/pull/826

This pull request introduces documentation updates to add support for SubWallet, a non-custodial wallet solution, in the context of Polkadot Hub. The updates provide detailed instructions for setting up and using SubWallet to interact with Polkadot Hub's Smart Contract module.

### Documentation Updates:

* Added a new section for SubWallet in `develop/smart-contracts/wallets.md` and `llms.txt`, explaining its features and compatibility with Polkadot, Substrate, and Ethereum ecosystems. The section includes step-by-step instructions for installing and configuring the SubWallet browser extension to connect to Polkadot Hub TestNet. [[1]](diffhunk://#diff-83f55b5463c4692f9eea61ef1866621384a5dfe5b2691f7dbbe0bdf357d44a80R42-R71) [[2]](diffhunk://#diff-f43a9e2f7da89777ca234fc2f236252b27a44652586c9514b243e81bf1b19114R12703-R12732)
* Included visual aids, such as screenshots and links to relevant resources, to enhance user understanding of the setup process. [[1]](diffhunk://#diff-83f55b5463c4692f9eea61ef1866621384a5dfe5b2691f7dbbe0bdf357d44a80R42-R71) [[2]](diffhunk://#diff-f43a9e2f7da89777ca234fc2f236252b27a44652586c9514b243e81bf1b19114R12703-R12732)

PR #837, opened 2025-07-23
Remove .DS_Store files

brunopgalvao:
Removes occurrences of the `.DS_Store` file that got in from this PR: https://github.com/polkadot-developers/polkadot-docs/pull/832


eshaben (commented):
Thank you! We should add any .DS_Store file to the .gitignore and also go back and remove these cached instances so they cannot be committed again (`git rm --cached`)

nhussein11 (approved):

eshaben (approved):
Thanks!

PR #774, opened 2025-07-07
Add content around working with non-sufficient assets

brunopgalvao:
Resolves https://github.com/polkadot-developers/polkadot-docs/issues/698

This pull request includes:
- Expanation around what sufficient and non-sufficient assets are, the UX that is available for them on Asset Hub, and what to consider when working with non-sufficients
- Best practices for teleporting assets; sufficients and non-sufficients
- Example of a batch teleport with Dry Run
- Cisco's [Dry Run example](https://hackmd.io/@n9QBuDYOQXG-nWCBrwx8YQ/SJ-rbyFkee)

Created a follow-up comment here: https://github.com/polkadot-developers/polkadot-docs/issues/813#issuecomment-3134173197

Follow-up PRs:
- https://github.com/polkadot-developers/polkadot-docs/pull/477
- https://github.com/polkadot-developers/polkadot-docs/pull/887

PR #849, opened 2025-07-25
Add @paraspell/sdk

brunopgalvao:
This pull request introduces a new dependency to the project by adding the `paraspell_sdk` package to the `variables.yml` file.

* Dependency management:
  * Added `paraspell_sdk` with version `10.10.11` to the `dependencies` section in `variables.yml`.

### DIFF
--- a/variables.yml
+++ b/variables.yml
@@ -82,4 +82,7 @@ dependencies:
     ethersjs:
       name: '@ethersjs'
       version: 6.13.5
+    paraspell_sdk:
+      name: '@paraspell/sdk'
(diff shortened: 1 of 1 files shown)


nhussein11 (approved):
Remember that if we merge this PR, the non sufficient asset pr should be updated to use - `paraspell_sdk.name`, here for example https://github.com/polkadot-developers/polkadot-docs/pull/774/files#diff-e95111afa4706333a90d4be36f56f6c35796c05374e9b4467afbde56be9c1e9cR91

PR #855, opened 2025-07-29
Update offenses

filippoweb3:
See Stage 4 https://github.com/paritytech/polkadot-sdk/issues/4359


filippoweb3:
This table is identical to the one on the Polkadot Wiki. To avoid maintaining it in two places and reduce the risk of inconsistencies, I suggest we remove it from the docs and keep the canonical version on the Wiki. The Wiki is also more likely to be updated promptly as changes occur.

Let me know if that works for everyone.

nhussein11:
> This table is identical to the one on the Polkadot Wiki. To avoid maintaining it in two places and reduce the risk of inconsistencies, I suggest we remove it from the docs and keep the canonical version on the Wiki. The Wiki is also more likely to be updated promptly as changes occur.
> 
> Let me know if that works for everyone.

@filippoweb3 I agree on this, and we can add a brief note in the docs directing to the wiki. In that case, which link should we use?

nhussein11:
Closing this PR in favor of https://github.com/polkadot-developers/polkadot-docs/pull/863/files#diff-b940acc91c45c37030d336c9fc96390ab562314def035b2a5f3c262f10f5fd5dR22-R27

PR #878, opened 2025-08-05
add trigger

Telucero:
adds the action trigger


albertov19:
@Telucero let's hold this until we have proper QA done with other docs first :)

PR #900, opened 2025-08-13
Explain that HRMP channels take a session boundary to open

seadanda:
It was pointed out to me in DM that our docs to open a HRMP channel with a system parachain don't make it clear that the channel doesn't actually open until the next session.

This should make it clear without changing too much.


brunopgalvao:
Closing this PR in favor of https://github.com/polkadot-developers/polkadot-docs/pull/907

PR #898, opened 2025-08-12
Add ParaSpell XCM SDK Guide

brunopgalvao:
This pull request introduces documentation and code samples for the ParaSpell XCM SDK, a toolkit designed to simplify cross-chain development on Polkadot. It adds a new ParaSpell section to the interoperability docs, provides an overview of the SDK and related tools, and includes practical code examples and their outputs to help developers understand how to use the SDK for asset transfers, dry runs, ED verification, and transfer info retrieval.

**Documentation Structure and Content**

* Added a new ParaSpell section to the interoperability navigation and created a dedicated navigation file for the SDK documentation. (`develop/toolkit/interoperability/.nav.yml`, `develop/toolkit/interoperability/paraspell/.nav.yml`) [[1]](diffhunk://#diff-05a3493ff03405bff6b5fc09e7634878013395a929b6f5095635abed7af6a1dfR6) [[2]](diffhunk://#diff-cb988cabbba8d9887756f9e93376783f114c41977940db0bb977932bd79f3f03R1-R5)
* Introduced an index page with a comprehensive overview of ParaSpell, its goals, and its suite of XCM tools, including SDK, API, Router, Analyser, Visualizator, and Playground. (`develop/toolkit/interoperability/paraspell/index.md`)

**Code Samples and Output Snippets**

* Added a full example script demonstrating how to use the ParaSpell SDK for asset teleportation, dry run, existential deposit (ED) verification, and transfer info retrieval. (`.snippets/code/develop/toolkit/interoperability/paraspell/index.ts`)
[...]

PR #941, opened 2025-08-28
Revert "Add ParaSpell XCM SDK Guide"

joepetrowski:
Reverts polkadot-developers/polkadot-docs#898

This does not make any sense. There is no reason to teleport to BH.


brunopgalvao:
@joepetrowski would I good first user journey for developers who want to learn how to do an XCM transfer using ParaSpell be: "Send DOT from AH to People" or do you suggest something more comprehensive such "Send DOT from AH to People and create an identity on People chain".

PR #959, opened 2025-09-02
Update hardhat.md

ottovlotto:
Hey team, this only works with v22 or up (I double checked with Beto who wrote the plugin) so can we quickly update this as some "low-hanging fruit" here? :) :)

### DIFF
--- a/develop/smart-contracts/dev-environments/hardhat.md
+++ b/develop/smart-contracts/dev-environments/hardhat.md
@@ -30,7 +30,7 @@ Hardhat is a robust development environment for Ethereum-compatible chains that 
 
 Before getting started, ensure you have:
 
-- [Node.js](https://nodejs.org/){target=\_blank} (v16.0.0 or later) and npm installed.
+- [Node.js](https://nodejs.org/){target=\_blank} (v22.0.0 or later) and npm installed.
 - Basic understanding of Solidity programming.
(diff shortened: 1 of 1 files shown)


eshaben (approved):

nhussein11 (commented):
Thank you @ottovlotto! Quick question: Have you followed the entire guide using `node 22` and the `hardhat-polkadot` version that the guide uses (hardhat-polkadot@0.1.9)?

Just checking to make sure this change won't cause any issues with some steps in the guide.

ottovlotto:
heya @nhussein11 I didn't but i can do tomorrow to be safe before you action the change? :)

ottovlotto:
heya @nhussein11 sorry I still haven't had a chance to go through all these and in the interests of moving forward I had a thought - it already says at the bottom of the docs "Consider using [Node.js](https://nodejs.org/) 22.18+ and [npm](https://www.npmjs.com/) version 10.9.0+ to avoid issues with the plugin" - I'm guessing someone checked it worked then? 

Also if we're not comfortable totally changing it without doing a thorough check (fair), could we move that disclaimer up further so it's super clear? So like this:
"Prerequisites
- [Node.js](https://nodejs.org/) (v16.0.0 or later) and npm installed. Consider using [Node.js](https://nodejs.org/) 22.18+ and [npm](https://www.npmjs.com/) version 10.9.0+ to avoid issues with the Polkadot plugin" ? :)

PR #1007, opened 2025-09-19
Latest Node recommended, fixed indentation

ottovlotto:
Breaking change with latest Hardhat plugin req Node 22.5, also fixed indentation :)


ottovlotto:
@nhussein11 ANOTHER node version update as of this week 😅 but means I can fix up the indentation while I am at it at least!

nhussein11 (approved):
Hey @ottovlotto, thank you! I’m starting to think that we could create a bot or something to automate these kinds of updates. Not sure if we have the bandwidth for that right now, but I’m just thinking out loud as a future idea.

Approved! Thanks!

PR #1180, opened 2025-11-05
removes deprecated files related to legacy vale workflow

dawnkelly09:
## 📝 Description

Removes the long-deprecated vale workflow file


eshaben (approved):
Thanks!

PR #1232, opened 2025-11-13
Remove foundry and ethers

eshaben:
## 📝 Description

This PR removes all of the Foundry pages for the MVP release. The ones that we've spent time on updating are just hidden til future notice, and the other ones we haven't started on, I just deleted.

It also removes the Ethers.js guides under Deploy a * Contract sections, not to be added back later.


dawnkelly09 (approved):
🚢

PR #1331, opened 2025-12-15
redesign changes

eshaben:
## 📝 Description

This PR updates the mermaid hardcoded styles so we just default to the ones defined in mkdocs. Goes with https://github.com/papermoonio/polkadot-mkdocs/pull/194


0xlukem (approved):
🚢  ❤️

PR #1651, opened 2026-04-22
Skill Generation Workflow: Generated files

aljosamakevic:
Auto-generated by the monitor-skills workflow.

This PR bundles all changes produced by the skill generation pipeline:

skill_coverage.json — page coverage state
skill_candidates.json — triage backlog (80 candidates classified + built; 45 supplementary; 12 not_applicable)
ref_repo_change_alerts.json — upstream reference repo alerts
repo_state.json — tracked reference repo commit SHAs
agent_skills_config.json — 55 generated skills consumed by the mkdocs plugin
Seeded initially from alt testing to skip re-triage/re-generation work. The workflow has since run the full triage + generate loop, so all classified candidates are now built.


agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
+{
+  "schema_version": "0.1",
+  "generated": "2026-04-21T23:59:00Z",
+  "content_hash": "sha256:b9fcc356e7673fb77911efa114edd9d68661b5a6d30ba7c7ee3753c0070e1920",
+  "project": {
+    "id": "polkadot-docs",
+    "name": "Polkadot Developer Docs",
+    "docs_base_url": "https://docs.polkadot.com"
+  },
```
eshaben: On the plugin side, we only consume `project.name`. We could probably get rid of the other two items. Not sure we even need any of these TBH, because the `llms_config.json` defines basic items like these and maybe that should be the source of truth for project-related details.

agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
...
+  "skills": [
+    {
+      "id": "create-polkadot-account",
+      "title": "Create a Polkadot Account Programmatically",
+      "objective": "Generate a new SR25519 Polkadot account including a 12-word mnemonic phrase and SS58 address using TypeScript.",
```
eshaben: We don't need the objective because the description should ideally be used instead as it's more useful since it includes the objective, trigger phases, and expected outputs.

agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
...
+          "Node.js v18+",
+          "npm or pnpm"
+        ]
+      },
+      "env_vars": [],
+      "steps": [
+        {
+          "order": 1,
+          "action": "Initialize the project",
+          "working_directory": ".",
```
eshaben, line 30: So we're not currently using this field at all, but we probably should. 

I was talking with Taylor, it might be helpful for an agent to outline the project structure so the agent understands the overall directory structure it's building toward. 

Something like a `## Project Structure` section near the top of the skill file (before the steps) would give the agent a target layout to work toward.

agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
...
+          "commands": [
+            "npx tsx create-account.ts"
+          ],
+          "expected_output": "Address: <SS58-encoded Polkadot address>\nMnemonic: <12-word BIP39 phrase>",
+          "description": "Execute the script. It prints a new SS58 address and 12-word mnemonic to the console. Record both — the mnemonic is required to sign transactions with this account and cannot be recovered if lost. Never share the mnemonic."
+        }
+      ],
+      "reference_code": {
+        "repo": "polkadot-cookbook",
+        "base_path": "polkadot-docs/chain-interactions/create-account",
```
eshaben: This base_path is pretty useless without a base URL. It's possible this used to be wired up using the `reference_repos` config? But now that no longer exists, and honestly, probably doesn't need to. I think it's fine to just define the entire URL here. OR at the very least, the repo config should be `<owner>/<repo_name>` instead of just `<repo_name>` that way we could stitch together the URLs.

Depending on the repo though, we may need to target `master` or `main` branches or if we're using versioning then we may need to target specific versioned releases. So I think this needs some thought

agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
...
+          "cause": "Dependencies not installed.",
+          "resolution": "Run 'npm install @polkadot/util-crypto @polkadot/keyring' in the account-creator directory."
+        }
+      ],
+      "supplementary_context": {
+        "description": "Load these pages when the user asks about next steps after account creation or about account concepts.",
+        "pages": [
+          {
+            "slug": "chain-interactions-send-transactions-with-sdks",
+            "url": "https://docs.polkadot.com/ai/pages/chain-interactions-send-transactions-with-sdks.md",
```
eshaben: This is outdated. The resolved markdown pages are now located at the same place as the server HTML file.

So in this example: `https://docs.polkadot.com/ai/pages/chain-interactions-send-transactions-with-sdks.md` should be `https://docs.polkadot.com/chain-interactions/send-transactions/with-sdks.md`

agent_skills_config.json
```diff
@@ -0,0 +1,12046 @@
...
+          "pattern": "Error: Cannot find module '@polkadot/util-crypto'",
+          "cause": "Dependencies not installed.",
+          "resolution": "Run 'npm install @polkadot/util-crypto @polkadot/keyring' in the account-creator directory."
+        }
+      ],
+      "supplementary_context": {
+        "description": "Load these pages when the user asks about next steps after account creation or about account concepts.",
+        "pages": [
+          {
+            "slug": "chain-interactions-send-transactions-with-sdks",
```
eshaben, line 97: I don't think this is really useful. Right now we use it in this context:

```
## Supplementary Context

Load these pages when the user asks about next steps after account creation or about account concepts.

- [chain-interactions-send-transactions-with-sdks](https://docs.polkadot.com/ai/pages/chain-interactions-send-transactions-with-sdks.md) — How to use the newly created account to sign and submit transactions on Polkadot Hub.
- [chain-interactions-accounts-query-accounts](https://docs.polkadot.com/ai/pages/chain-interactions-accounts-query-accounts.md) — How to query the balance and account state of the newly created account.
```

But are these slugs as titles really the best way to define these? Some feedback from Claude:

> The one thing that stands out is the slug as link text — chain-interactions-send-transactions-with-sdks isn't very readable. Since the URLs already contain that information, it would be cleaner to use a human-readable label if one is available in the config

ref_repo_change_alerts.json
```diff
@@ -0,0 +1,31 @@
+{
+  "schema_version": "1",
+  "generated": "2026-04-21T14:42:33Z",
+  "summary": {
+    "reference_changes": 0,
+    "skills_affected": 0
+  },
+  "alerts": [],
+  "history": [
```
eshaben: So a new entry gets added here every time an alert is emitted? Is this necessary? At what point do we cut the history off so this doesn't become a massive file?

repo_state.json
```diff
@@ -0,0 +1,15 @@
+{
+  "schema_version": "1",
+  "last_checked": "2026-04-21T14:42:33Z",
+  "docs_repo": {},
+  "reference_repos": {
```
eshaben, line 8: Ahhhh ok... curious why this was split out?

repo_state.json
```diff
@@ -0,0 +1,15 @@
+{
+  "schema_version": "1",
+  "last_checked": "2026-04-21T14:42:33Z",
+  "docs_repo": {},
```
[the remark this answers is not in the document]
eshaben: Curious why we need this if it's empty?

skill_candidates.json
eshaben, line 1: Are all of these fields used?

repo_state.json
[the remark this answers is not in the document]
aljosamakevic: The empty docs_repo was not intentional. The field IS tracked by the automation (monitor_ref_repo_changes.py) it just hadn't been populated yet in the alt- worfklow I trigerred. It SHOULD be populated and kept up to date when the worfkflow actually runs here on prod.

Filled it in with the actual fork point: master commit cc6c24e (2026-04-06), which is when I forked the polkadot-docs repo.

repo_state.json
aljosamakevic: If you mean why reference_repos used to be in agent_skills_config.json and now it's here - I split it to keep the config.json file focused on just skill definitions.

The general repo/project config stuff lives in registry.json (URLs/branches/raw_base_urls), and tracked state (like last-checked SHAs) live here in repo_state.json

agent_skills_config.json
aljosamakevic: Gotcha
I've updated agent-ready-docs-hub (in a sep PR) to not produce .id and .docs_base_url anymore in the config.json

agent_skills_config.json
aljosamakevic: > So we're not currently using this field at all, but we probably should.

You mean on the mkdocs-plugins render side?

> I was talking with Taylor...

I implemented Project Structure in the new agent-ready-docs-hub PR so new skills will have this, and stale skills will get this added to them when getting updated.

agent_skills_config.json
aljosamakevic: For now I'll do it like owner/name and add a branch field

agent_skills_config.json
aljosamakevic: Added a title field

ref_repo_change_alerts.json
aljosamakevic: History has a 30 item cap, but yes, it is unnecessary now. I made it as part of another thing that I removed a while ago, so I'll remove history too, you're right.

agent_skills_config.json
eshaben: I'm not using the name now either, so that can be removed at any time

agent_skills_config.json
eshaben: Yes, on the mkdocs-plugins side we're currently not using the `working_directory`

agent_skills_config.json
```diff
@@ -0,0 +1,12207 @@
...
+            "After calling dev_setStorage, send a dev_newBlock request with count: 1 to commit the change",
+            "If using Batch block build mode (default), storage changes are applied at next block production",
+            "Switch to Instant mode with dev_setBlockBuildMode to get automatic per-transaction block production during testing"
+          ],
+          "result": "Storage override committed after block production; account shows updated balance"
+        }
+      ]
+    },
+    {
+      "id": "build-dapp-viem-nextjs
```
eshaben: To make these links more useful, can we make changes to the hub so that we always render the full URL with the base `https://docs.polkadot.com/` prepended to it?

eshaben: The stuff in parenthesis is a command (for reference, I believe it's being pulled from this guide: https://docs.polkadot.com/parachains/launch-a-parachain/set-up-the-parachain-template/#prerequisites) but who knows

eshaben: Anyways, commands referenced inline in prose should be wrapped in backticks for clarity, but in a structured data context like skill.steps[n].commands where the schema already implies they're commands, the backtick formatting is redundant and can be handled in the mkdocs-plugins logic instead

eshaben: Some of this should be wrapped in a backtick with the programming language preserved so it's like:

\```toml
[dependencies]
...
pallet-collective = { version = "<SDK_VERSION>", default-features = false }
pallet-membership = { version = "<SDK_VERSION>", default-features = false }

Replace <SDK_VERSION> with the version matching your other SDK pallets (check existing entries). Add both to the std feature list:

[features]
std = [
    ...
    "pallet-collective/std",
    "pallet-membership/std",
]
\```

agent_skills_config.json
```diff
@@ -0,0 +1,12207 @@
...
+            "npm install --save-dev typescript tsx"
+          ],
+          "description": "Install @polkadot/util-crypto (WASM-backed mnemonic generation and cryptoWaitReady), @polkadot/keyring (SR25519 keyring), and tsx (TypeScript executor)."
+        },
+        {
+          "order": 3,
+          "action": "Fetch the account creation script",
+          "working_directory": "account-creator",
+          "reference_file": "create-account.ts",
+          "description": "Fetch the reference file and save it as 'create-account.ts'. No substitutions needed — the script generates a fresh random account on each execution. If the AI fetch tool is blocked, run: curl -sL https://raw.githubusercontent.com/polkadot-developers/polkadot-cookbook/main/polkadot-docs/chain-interactions/create-account/create-account.ts -o create-account.ts"
```
eshaben: ah haha this link isn't even real. It was hallucinated?

eshaben: This is the related test link in the cookbook repo: https://github.com/polkadot-developers/polkadot-cookbook/blob/master/polkadot-docs/chain-interactions/create-account/tests/docs.test.ts

But that's not what we would even want to add here

agent_skills_config.json
eshaben: This isn't even correct. Because here it's `main` but it should be `master`, but also because the raw_base_url is missing `refs/head`, should be:

`https://raw.githubusercontent.com/polkadot-developers/polkadot-cookbook/refs/heads/master`...

eshaben: I think we can probably just get rid of this

skill_candidates.json
aljosamakevic: yes

agent_skills_config.json
aljosamakevic: My agent says it better than I can: Codified in `step-writing-guide.md` ([papermoonio/agent-ready-docs-hub#4](https://github.com/papermoonio/agent-ready-docs-hub/pull/4), commit 431cd6d): prose fields → backticks for inline commands; `steps[].commands[]` stays bare since the mkdocs plugin handles rendering.

Fixed the flagged line in `add-existing-pallet-to-runtime` (commit 4cc9b09 here) — `rustup target add ...` is now backticked, content_hash recomputed. Other existing skills in the seed have similar bare-command-in-prose cases; rather than batch-rewriting 80 skills in this PR, they'll come into compliance as they're regenerated/updated.

agent_skills_config.json
aljosamakevic: This opened up a whole can of worms hahah

Two fixes:

Schema rules ([papermoonio/agent-ready-docs-hub@d8752eb](https://github.com/papermoonio/agent-ready-docs-hub/commit/d8752eb)) — banned curl-fallback URLs in step descriptions (the hallucination surface), and the generator now has to HTTP-verify each constructed reference URL before writing it.
Re-pointed 14 skills (commit on this branch) — reference_code.repo now points at polkadot-developers/polkadot-docs with base_path = .snippets/code/<page-path>. For 2 setup-only skills with no code includes, repo: "none". Removed all the hallucinated curl URLs; preserved the legitimate ones (Homebrew, AcalaNetwork chopsticks, etc.).
Result: 43/44 raw URLs in the file now HTTP 200; the 1 remaining is a parameterized SDK_VERSION placeholder for the agent to substitute at runtime.

eshaben: i love cans of worms 🪱

eshaben: can you clarify what you mean by "placeholder for the agent to substitute at runtime" - are you talking about one of the variables defined in `variables.yml`? Or is this something different?

agent_skills_config.json
eshaben: Ok, cool. So wait what files were regenerated so far?

aljosamakevic: I updated all of them based on all the new rules we put in place this past week, but they're still made from the forked version of the repo - so whichever content updates happened between the time I forked the repo and now - those have not gone through the workflow and have not been implemented into the generated skills

agent_skills_config.json
aljosamakevic: Yeah, but I remember after I changed something about that specific URL situation so that in the skill it's defined where to fetch the new sdk version from, so that way the url can be assembled right away

agent_skills_config.json
```diff
@@ -0,0 +1,12321 @@
...
+      ],
+      "steps": [
+        {
+          "order": 1,
+          "action": "Clone the reference repository",
+          "working_directory": ".",
+          "commands": [
+            "git clone https://github.com/polkadot-developers/revm-hardhat-examples.git"
+          ],
+          "description": "Clone revm-hardhat-examples. This repo contains the complete zero-t
```
eshaben, line 4449: lib.rs code is not in backticks. I imagine it's supposed to look like this:

\```rust
#[cfg(test)] mod mock;
#[cfg(test)] mod tests;
\```

eshaben, line 4416: I feel like these could just cause confusion. no links. what is pallets/pallet-custom? There is a link to the mock guide in the supplementary section, which is kind of interesting though that it wouldn't just be here as it's a prereq.

agent_skills_config.json
```diff
@@ -0,0 +1,12321 @@
...
+        },
+        {
+          "order": 2,
+          "action": "Create or download a fork configuration file",
+          "working_directory": ".",
+          "description": "Create a YAML configuration file for the parachain to fork. Pre-built configs for common Polkadot ecosystem chains are available at https://github.com/AcalaNetwork/chopsticks/tree/master/configs. Download one with:\n  curl -sL https://raw.githubusercontent.com/AcalaNetwork/chopsticks/master/configs/moonbeam.yml -o parachain.yml\n\nOr write one manually. Minimal example:\n  endpoint: wss://moonbeam.public.blastapi.io\n  mock-signature-host: true\n\nOptional: override storage in the config to pre-fund test accounts:\n  import-storage:\n    System:\n      Account:\n        - - - 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY  # Alice address\n          - data:\n              free: '100000000000000'\n\nReplace the endpoint with the WSS URL for your target parachain. All CLI config options (genesis, timestamp, endpoint, block, db, port, build-block-mode, import-storage, allow-unresolved-imports, html, mock-signature-host) can be set in this file or passed as CLI flags."
```
eshaben: I feel like we got rid of these types of curl commands but maybe not. Either way, the curl command isn't wrapped in back ticks. Neither are the yaml examples

agent_skills_config.json
eshaben: Seems like the single backtick issue was fixed maybe in one place, but not here even though all the files have been updated?

eshaben: I feel like there should be a link to the create-frame-pallet skill, otherwise if someone just downloads this file, I think this is going to be kind of useless

eshaben: another example of missing backticks

eshaben (commented):
spot checked about 10ish files, seeing mostly the same things as mentioned here

eshaben (approved):
If the testing goes well Aki, I think we can release this as it is and make a follow up PR that addresses some of the items in my last review. Approving now

eshaben:
@aljosamakevic please mark the comments as resolved if they've been addressed. again, just going to approve this and then we can continue to audit these files

eshaben (approved):

aljosamakevic:
> @aljosamakevic please mark the comments as resolved if they've been addressed. again, just going to approve this and then we can continue to audit these files

I kept the comments as is on purpose, because my commits were/are related to other things

0xlukem (approved):
LGTM

PR #1728, opened 2026-07-09
Remove dead chain-snapshot link from validator requirements

ReinhardHatko:
The `stakeworld.github.io/docs/dbsize` link in the validator storage requirements is offline (404). StakeWorld moved to `stakeworld.io`, but the database-sizes page wasn't carried over (the site is an SPA that returns 200 for any path, so a link there would falsely pass the checker while showing an empty page). The link is a supplementary "for current estimates" pointer; the storage guidance is complete without it, so this removes it.


UtkarshBhardwaj007 (approved):
LGTM

PR #1725, opened 2026-07-02
Remove the Kapa AI chatbot widget

ReinhardHatko:
Removes the embedded Kapa AI chatbot widget from the theme (preconnect, widget script, and the shadow-DOM style-patch script). Verified: `mkdocs build --strict` passes and no Kapa references remain in the output.

PR #1730, opened 2026-07-10
Add docs assistant integration (widget + corpus feed)

illegalcall:
Replaces the kapa.ai widget (removed in #1725) with a self-hosted assistant.

The part that actually answers questions is a separate service in its own repo, [paritytech/polkadot-docs-assistant](https://github.com/paritytech/polkadot-docs-assistant). This PR is just the docs-side half: the widget, and the build step that publishes the docs in a form the service can read. The only thing the two share is that published file.

What's in here:

- `generator/generate_feed.py` and `hooks/ai_feed.py` do what the papermoon `ai_docs` plugin used to do. At build time they write `ai/llms-full.jsonl` (the docs split into chunks), a `.md` copy of every page, `llms.txt`, and the "Markdown for LLMs" dropdown. The chunking lives in the generator; the hook just wires it into the mkdocs build and respects `exclude_docs`.
- `docs/js/docs-assistant.js` and its CSS are the chat widget. It reads its API URL from `DOCS_ASSISTANT_API_URL` at build time and renders nothing if that's empty.
- `ai-feed-checks.yml` regenerates the feed on docs PRs and checks the schema, the page/chunk counts, and that no snippet includes leaked into the output.

I dropped the `ai_docs` plugin from `mkdocs.yml` but left `papermoon-mkdocs-plugins` in `requirements.txt`, since `link_processor`, `minify` and `page_toggle` still use it.

Heads up: the widget stays hidden until `DOCS_ASSISTANT_API_URL` is set in the deploy config. Merging this on its own only changes the generated feed and per-page markdown.

### DIFF
--- a/.github/workflows/check-mkdocs-build.yml
+++ b/.github/workflows/check-mkdocs-build.yml
@@ -34,3 +34,28 @@ jobs:
 
       - name: Build MkDocs site
         run: mkdocs build -d site
+
+      - name: Check the AI corpus feed
+        # Validates the real deploy artifact — the feed hooks/ai_feed.py just wrote via
+        # the build above — so this gate can't diverge from what actually ships (same
+        # page set, resolved snippets, and macros env as production).
+        run: |
+          python - <<'EOF'
+          import json, re
+          path = 'site/ai/llms-full.jsonl'
+          raw = open(path, encoding='utf-8').read()
+          rows = [json.loads(l) for l in raw.splitlines() if l.strip()]
+          pages = {r['page_id'] for r in rows}
+          assert len(rows) > 1500, f"chunk count collapsed: {len(rows)}"
+          assert len(pages) > 180, f"page count collapsed: {len(pages)}"
+          required = {"source", "page_id", "url", "anchor", "page_version_hash",
+                      "page_title", "title", "text"}
+          missing = required - set(rows[0])
+          assert not missing, f"schema regression, missing: {missing}"
+          assert '--8<--' not in raw, "unresolved --8<-- snippet directives in feed"
+          # leaked mkdocs template filters (e.g. {{ x | fix_url }}); targets the pipe so
+          # legit literals like {{ $labels.instance }} don't false-positive.
+          leaked = re.findall(r'\{\{[^}\n]*\|[^}\n]*\}\}', raw)
+          assert not leaked, f"unresolved template filters leaked into feed: {leaked[:3]}"
+          print(f"feed OK: {len(rows)} chunks / {len(pages)} pages, schema intact, no raw snippets or template leaks")
+          EOF
--- a/docs/js/docs-assistant.js
+++ b/docs/js/docs-assistant.js
@@ -1,0 +1,227 @@
+/* Polkadot Docs Assistant widget.
+   Self-hosted chat over the docs corpus — replaces the former kapa.ai embed.
+   Talks to the assistant API (see the service repo): POST /chat, POST /feedback.
+
+   Configured from the script tag in main.html:
+     <script defer src="js/docs-assistant.js" data-api-url="https://..."></script>
+   If data-api-url is empty the widget renders nothing (dark launch).
+*/
+
+(function () {
+  'use strict';
+
+  if (typeof window === 'undefined') return;
+
+  const script = document.currentScript;
+  const API_BASE = (script && script.dataset.apiUrl ? script.dataset.apiUrl : '').replace(/\/$/, '');
+  if (!API_BASE) return; // no endpoint configured — ship dark
+
+  const HISTORY_TURNS = 6; // question+answer pairs sent back for follow-up rewriting
+  const history = [];
+  const sessionId = 'da-' + Date.now().toString(36) + '-' + Math.random().toString(36).slice(2, 8);
+
+  // ---------- tiny markdown renderer (no dependencies) ----------
+  function escapeHtml(s) {
+    // Must escape quotes too: the answer is untrusted LLM output rendered via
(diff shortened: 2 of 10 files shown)


EnderOfWorlds007:
@ReinhardHatko can you assist?

ReinhardHatko (commented):
Really nice PR — thanks for the care here. Dropping the papermoon plugin dependency, keeping generation in-repo with a single shared `process_page`/`chunk_row`, shipping the widget dark until an endpoint is set, and adding a feed-integrity gate is more discipline than most feature PRs. The LLM-output escaping in particular is exemplary. Overall I'd rate it **approve-with-changes**: one security gap to fix before merge, plus one systemic note about what the CI gate actually validates.

### 🔴 Fix before merge

**1. `docs/js/docs-assistant.js` — the source-chip `href` skips the scheme check the answer links get.** The `mdInline` link builder deliberately enforces `https?:` to block `javascript:`/`data:` injection, but the source chips do `chip.href = s.url` with no validation. `s.url` crosses a trust boundary (the API response), and the file's own `escapeHtml` comment already treats that output as untrusted — so this is a gap in the PR's own threat model. Low real-world likelihood (sources are normally `docs.polkadot.com` URLs), but cheap to close:

```js
if (s.url && /^https?:$/i.test(new URL(s.url, location.href).protocol)) {
  chip = el('a', 'da-chip', s.page_title || s.title || pid);
  chip.href = s.url; chip.target = '_blank'; chip.rel = 'noopener';
} else {
  chip = el('span', 'da-chip', s.page_title || s.title || pid);
}
```

### 🟠 Worth addressing — the gate validates a different path than deploys

The "CI feed and deployed feed can't diverge" guarantee holds for chunk *shape*, but not for page *selection* or resolved *text*:

- **Page set:** CI runs the standalone generator (`os.walk` + `llms_config.json`); deploy runs the hook (mkdocs `exclude_docs` file list + per-page `hide_ai_actions`). A page excluded in `mkdocs.yml`/`hide_ai_actions` but not in `llms_config.json` is green in CI yet missing from production (and vice-versa). Effectively three exclusion lists to keep in lockstep.
- **Snippets:** CI is offline (`GEN_SNIPPET_URL_DOWNLOAD=0`, drops remote `--8<--`); the hook fetches them. For pages with remote snippets the `text`, chunk boundaries, and `page_version_hash` differ — the gate can't validate exactly those pages.
- **Template vars:** `resolve_vars` swallows exceptions and returns raw text, and the bare `Environment()` doesn't know the site's mkdocs-macros filters, so `{{ x | fix_url }}` can leak as literal `{{ … }}` into the feed. The gate guards leaked `--8<--` but not leaked `{{ }}`.

None crash — they let a green gate ship a subtly-wrong corpus, which is the one thing the gate exists to prevent. Ideally CI exercises the deploy (hook) path, or at least shares the same exclusion inputs + one online snippet mode.

### 🟡 Nits (non-blocking)

- **`docs-assistant.js` markdown renderer** (hand-rolled, so the usual edge cases): single-`*` emphasis eats unrelated asterisks (`2 * 3 = 6 and *note*` → stray `<em>`); an unbalanced code fence swallows the rest of the answer into one `<pre>`; and the link regex `(https?:[^)\s]+)` truncates URLs containing `)` (e.g. `..._(programming_language)`). Fixing the emphasis one is the highest-value; longer term a tiny vetted renderer would remove the class.
- **`ai-feed-checks.yml`** thresholds (`> 1500` chunks / `> 180` pages) are absolute floors — a legit large removal red-fails, and a ~50% collapse still passes. Consider basing them on the base-branch counts.
- Stale `ai_docs`-plugin comments + an orphaned `data-variant` in `material-overrides/partials/page-badges.html` (the hook now matches the `page-header-row` class, not that attribute).

### ✅ Checked and cleared
- No duplicate `{% block libs %}` (master has 0 after the kapa removal, this PR adds 1) — build is fine on that front.
- `_git_dates()` newest-per-file parsing is correct.

Altitude-wise, the in-house reimplementation is the right call given the goal is to drop the papermoon dependency — not over-engineering. Nice work overall.

illegalcall:
Thanks — really thorough review. All addressed:

- **🔴 Source-chip href** — added an `httpUrl()` guard; a source only becomes a link if it's `http(s)`, matching the scheme check `mdInline` already does for answer-body links.
- **🟠 CI vs deploy path** — moved the feed check onto the **real build artifact** rather than the offline generator. `check-mkdocs-build.yml` already runs `mkdocs build` on every PR, so it now asserts schema / counts / no-`--8<--` / no-leaked-filters on `site/ai/llms-full.jsonl` — same page set, resolved snippets, and macros env as production. Deleted the standalone `ai-feed-checks.yml`. The template-leak guard targets filter exprs (`{{ x | fix_url }}`) so legit literals like `{{ $labels.instance }}` don't false-positive.
- **🟡 Emphasis** — fixed; `2 * 3` no longer italicizes (lookbehind-free, so no issue on older Safari).
- **🟡 Link parens** — fixed; `..._(programming_language)` URLs no longer truncate at the first `)`.
- **🟡 Stale `ai_docs` comment** — reworded in `page-badges.html`.

Two I'd push back on:
- **`data-variant` isn't orphaned** — it's consumed by `toggle-pages.js` and `toggle-pages.css` (`.page-header-row[data-variant]`), so I kept the attribute and only fixed the stale comment.
- **Unclosed code fence** — extending to end-of-input is CommonMark-correct, so I left it (agree the hand-rolled renderer is worth swapping for a vetted one eventually).

Kept the absolute count thresholds for now — base-relative would mean building the base branch too; happy to revisit.

ReinhardHatko (approved):
LGTM

PR #1732, opened 2026-07-14
Scroll to the cited section in place from source chips

illegalcall:
Makes the source chips navigate smarter:

- **Cited section is on the page you're already reading** → clicking the chip smooth-scrolls to that heading **in place** — no reload, and the chat panel stays open.
- **Cited section is on another page** → opens that page at the exact section in a new tab (so the conversation isn't lost), as before.

The chips already deep-linked to the exact section — the feed's `url` includes the `#anchor` — so this only adds the in-place scroll for same-page sources. Verified the corpus anchors match the live heading ids (spot-checked several pages, 100%).

Not included: reload-free navigation to *other* pages. That needs MkDocs Material's site-wide instant navigation (`navigation.instant`), which changes navigation for the whole site and affects other custom scripts, so it belongs in its own tested change rather than riding along here.

### DIFF
--- a/docs/js/docs-assistant.js
+++ b/docs/js/docs-assistant.js
@@ -90,6 +90,15 @@
     catch (e) { return null; }
   }
 
+  // true if href points at the page the reader is already on (ignoring hash/trailing slash)
+  function samePage(href) {
+    try {
+      const u = new URL(href, location.href);
+      const strip = function (p) { return p.replace(/\/+$/, ''); };
+      return u.origin === location.origin && strip(u.pathname) === strip(location.pathname);
+    } catch (e) { return false; }
+  }
+
   const btn = el('button', 'da-launcher');
   btn.type = 'button';
   btn.setAttribute('aria-label', 'Open the docs assistant');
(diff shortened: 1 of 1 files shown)


ReinhardHatko (commented):
Nice, focused change — chips smooth-scroll to the cited heading in place and keep the chat open, with a clean fallback (anchor not found → let the browser handle the `#hash`). I confirmed the sticky-header offset is a non-issue: Material sets `scroll-margin-top` on anchor targets and `scrollIntoView` honors it, so it lands like native hash navigation. `build-docs` and link-validation are green. **Approving.**

One optional suggestion inline — it also makes the in-place scroll work in local/preview/subpath builds. Only `check-label` is red (needs the `B1 - Ready to Merge` label).

docs/js/docs-assistant.js
```diff
@@ -90,6 +90,15 @@
...
   }
 
+  // true if href points at the page the reader is already on (ignoring hash/trailing slash)
+  function samePage(href) {
+    try {
+      const u = new URL(href, location.href);
+      const strip = function (p) { return p.replace(/\/+$/, ''); };
+      return u.origin === location.origin && strip(u.pathname) === strip(location.pathname);
+    } catch (e) { return false; }
+  }
```
ReinhardHatko, line 100: **Optional enhancement.** Right now the in-place scroll only fires on production: this compares the cited (absolute, production) `s.url` against `location`, which only matches when the site is served from `docs.polkadot.com`. On localhost / preview / a `github.io` subpath the origins differ, so chips always open a new tab there.

Comparing against the page's **canonical** URL removes that limitation — Material emits the absolute production URL as `<link rel="canonical">` on every page (from `site_url`), regardless of where it's served, so same-page detection lines up everywhere and becomes testable locally. Falls back to `location` if a page has no canonical.

```suggestion
  // true if href points at the page the reader is on, independent of where the site is
  // served: compare against the page's canonical (production) URL, which Material emits
  // from site_url on every page — so this also works on localhost/preview/subpath deploys.
  function samePage(href) {
    try {
      const canon = document.querySelector('link[rel="canonical"]');
      const cur = new URL(canon ? canon.href : location.href, location.href);
      const u = new URL(href, cur);
      const strip = function (p) { return p.replace(/\/+$/, ''); };
      return u.origin === cur.origin && strip(u.pathname) === strip(cur.pathname);
    } catch (e) { return false; }
  }
```

PR #1740, opened 2026-08-19
Polkadot Apps update

bucanero:
## 📝 Description

This PR updates the Polkadot Apps section adding new content and fixes.

## ✅ Checklist

- [x] Changes tested  
- [x] [PaperMoon Style Guide](https://github.com/papermoonio/documentation-style-guide) followed


docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,214 @@
...
+
+!!! note "Contracts are optional"
+    Many Products never need a contract. If all you need is durable content or real-time state between users, [Store Data on Chain](/apps/build/store-data-on-chain/) and [Publish and Subscribe to Off-Chain Data](/apps/build/pub-sub-off-chain-data/) cover those without any contract at all. Reach for a contract when you need enforced, shared on-chain logic.
+
+## Prerequisites
+
+Before starting, ensure you have:
+
+- A Polkadot Product project running locally. See [Set Up Your Project](/apps/build/#set-up-your-project).
+- The [Polkadot App](/apps/) installed and paired, so you can sign the deploy on your phone. See [Install Desktop and Pair](/apps/get-started/).
```
illegalcall: Does `cdm` actually take the phone signer? I couldn't find any QR or pairing path in the cdm CLI. Looks like with no `--suri` it falls back to `//Alice`, and since contract names are first-writer-owns, that would park the name on the shared dev key. Should this mention `--suri` instead?

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,214 @@
...
+!!! tip
+    Pass `cdm setup --check` to verify the toolchain without installing anything.
+
+## Set the Target Network
+
+Open `cdm.json` and confirm the `registry` field points at the `ContractRegistry` for your target network. This address is network-specific; using the wrong one publishes your contract to the wrong registry. The manifest starts out with only your dependencies declared:
+
+```json title="cdm.json (before deploy)"
+{
+  "registry": "0xf62c2ece29cd8df2e10040ecfa5a894a5c5d9cb0",
```
illegalcall: Is `0xf62c2ece...` still live? `eth_getCode` comes back empty for it on paseo-next-v2. Also, does `cdm deploy` even read this field? Looked to me like it resolves from `-n` / `--registry-address`, so editing it might be a no-op. Worth dropping it?

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Build and Deploy
+
+Deploy the contract with `cdm deploy`, selecting your target network with `-n`. This builds the bytecode, deploys it to Asset Hub, uploads the ABI to the Bulletin Chain, and registers the name, all in one signed flow:
+
+```bash
+cdm deploy -n paseo
+```
+
+Sign the deploy with your phone when prompted. When it completes, `cdm` writes the real deployment details back into `cdm.json`, so the manifest now carries the address, ABI, and metadata CID your frontend needs:
```
illegalcall: Does `cdm` write back to `cdm.json` here? I only found `writeCdmJson` called from the install path. The `shared-counter` template also seems to ship a populated `contracts` map already, so does the manifest really start with just dependencies?

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,214 @@
...
+
+To depend on a contract someone else already published, add it to `cdm.json` and install it. `cdm install` resolves the name against the on-chain registry, fetches the ABI from the Bulletin Chain, and writes the address and ABI into your manifest:
+
+```bash
+cdm install @some-scope/their-contract -n paseo
+```
+
+This is the same manifest your own deploy produces, so the frontend integration below works identically whether you deployed the contract or installed it.
+
+## Call the Contract From Your Frontend
```
illegalcall, line 191: Should this section call `ensureContractAccountMapped` before the first write? `contracts.md` says every tx fails with `AccountNotMapped` until it runs, and `quick-start.md` points here for mapping. I think it's the app-scoped account that's unmapped, not the CLI login one.

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,214 @@
...
+  cdmJson,
+  client.raw.assetHub,
+  paseo_asset_hub,
+  { signerManager },
+);
+
+const counter = manager.getContract('@my-app/counter');
+
+// Read: a dry run against the best block. No signature, no fee.
+const { value } = await counter.getCount.query();
```
illegalcall: Should this check `.success` before reading `.value`? On a revert `.value` holds the dispatch-error payload, so this would render the error object as the count.

docs/apps/get-started/set-up-your-ai-agent.md
```diff
@@ -0,0 +1,123 @@
...
+```text
+You are helping me build a Polkadot Product: a sandboxed web app (HTML/JS/CSS)
+that runs inside a Polkadot Host (Polkadot Desktop, the Polkadot App, or Polkadot
+Web) and is addressed by a .dot name.
+
+Toolchain:
+- Build app code with @parity/product-sdk (TypeScript). Install its skills first.
+- Deploy with the playground CLI (`pg` / `playground`).
+- For smart contracts, use the Contract Dependency Manager (`cdm`); contracts are
+  Rust compiled to PolkaVM on pallet-revive (Asset Hub). Never target EVM/Solidity
```
illegalcall: Do we want to rule out Solidity here? cdm has `foundry-counter` and `hardhat-counter` templates, playground-cli has `hasBuildableSolidityProject` plus a `pr-deploy-foundry` CI cell, and our skills page says "PolkaVM / Solidity". Maybe allow both Rust and Solidity, and ban mainnet-EVM tooling instead?

docs/apps/get-started/set-up-your-ai-agent.md
```diff
@@ -0,0 +1,123 @@
...
+- https://github.com/paritytech/playground-cli (the deploy CLI)
+- https://github.com/paritytech/contract-dependency-manager (contracts)
+
+Rules to follow:
+- All chain reads, signing, and storage route through the Host via @parity/product-sdk.
+  Never open a direct WebSocket/RPC connection or dial a public IPFS gateway.
+- Sign with a product account: getProductAccount(...).getSigner(). Approvals happen
+  on the user's phone. On Paseo Next v2, do not use Polkadot.js-style signing (the
+  AsPgas signed extension breaks it) — use the product-account signer.
+- Fallible SDK calls return a Result; check `.ok` before reading `.value`.
```
illegalcall: Should this split by API? Contract `query` returns `{ success, value, gasRequired }` with no `.ok`, so `if (r.ok)` never passes and reads fail silently. `contracts.md` says "Check `.success`, not `.ok`". Since people paste this block into agents, worth spelling out both shapes?

docs/apps/get-started/set-up-your-ai-agent.md
```diff
@@ -0,0 +1,123 @@
...
+For other agents (Cursor, Codex, Cline, Windsurf, or a custom harness), clone `paritytech/product-sdk` and register the skills under `product-sdk/skills/` with your agent's context mechanism. The [Skills reference](/reference/apps/skills/) covers the full list, how to register them, and how to keep them current.
+
+!!! tip "Keep skills current"
+    Product surfaces are in active development. A stale skill teaches stale API patterns, which is worse than no skill. Refresh after major releases.
+
+## Start From a Template
+
+If you scaffold a Product from a Playground template — for example by cloning a moddable app with `pg mod` — the project already ships agent-instruction files, so your agent picks up the conventions automatically:
+
+- `CLAUDE.md` for Claude Code, `AGENTS.md` for Codex and other `AGENTS.md`-aware tools.
```
illegalcall, line 89: Is this true for any moddable app, or mainly the starter template? `playground-app-template` and `playground-tutorial` have all five, but `playground-app-community` only has `CLAUDE.md` and `playground-constellation` has none. Scope it to `pg init`?

docs/apps/get-started/set-up-your-ai-agent.md
```diff
@@ -0,0 +1,123 @@
...
+- `.clinerules`, `.windsurfrules`, and `.github/copilot-instructions.md` for Cline, Windsurf, and GitHub Copilot.
+
+Run the project's setup script (`./setup.sh`) once to install dependencies and fetch the latest skills into the project.
```
illegalcall: Do most moddable apps ship a `setup.sh`? The CLI skips it with "no setup.sh found" and the changelog says apps without one are the common case. I think `pg mod` also runs it when it exists, so does the reader need to?

docs/apps/product-sdk/tx.md
```diff
@@ -0,0 +1,103 @@
...
+- To submit several calls as one atomic batch through the Utility pallet (`batchSubmitAndWatch`).
+- For the surrounding steps: dry-run extraction and weight buffering, Asset Hub account mapping for `pallet-revive`, retries with backoff, and dev signers for tests.
+- Do not use it to build the transaction object or open connections; that is the typed API from the chain client. To submit a contract call, prefer the higher-level [Contracts](/apps/product-sdk/contracts/) package, which wraps this one.
+
+## Core Concepts
+
+- **`submitAndWatch(tx, signer, options)`**: Signs, broadcasts, and watches through `signing`, `broadcasting`, `in-block`, and `finalized`. It returns a `Result`, resolving expected failures on the error channel rather than throwing.
+- **`TxResult` and `TxStatus`**: `TxResult` carries the `txHash`, an `ok` flag for on-chain dispatch success, the `block`, and the emitted `events`. `TxStatus` drives the `onStatus` callback for progress UI.
+- **Two success checks**: `result.ok` means the transaction was included; `result.value.ok` means the on-chain dispatch itself succeeded. A transaction can land in a block and still fail at dispatch.
```
illegalcall: Can `result.value.ok` ever be false? Looks like `submitAndWatch` routes dispatch failures to `result.error` as `TxDispatchError`, so this branch may be unreachable. Same for the Limitations line at the bottom.

docs/apps/product-sdk/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+The [Product SDK](https://github.com/paritytech/product-sdk) is the TypeScript SDK for building Polkadot Products. It gives your Product typed access to everything the platform provides: chain reads, transaction signing, decentralized storage, off-chain messaging, smart contracts, and identity, all routed through the Host your Product runs inside.
+
+The SDK never dials an RPC endpoint itself. Every sensitive operation (signing, chain access, storage) goes through the Host, which selects the network, holds the user's keys, and prompts for approval on the user's phone. Your Product calls a typed method; the Host mediates the rest.
+
+Fallible operations return a typed `Result` instead of throwing, so you check `.ok` before reading `.value`. This pattern runs through the whole SDK and is covered in each [Build guide](/apps/build/).
+
+## Two Ways to Use the SDK
+
+The SDK ships as one umbrella package that re-exports everything, plus individual per-capability packages you can install on their own:
```
illegalcall: Is statement-store included in the umbrella? I didn't see a `./statement-store` subpath in its exports. And are the import paths identical? `@parity/product-sdk/cloud-storage` vs `@parity/product-sdk-cloud-storage` looked different to me.

illegalcall (commented):
Went through this fairly carefully. Structure and coverage are a big step up. Most of my notes are questions rather than blockers, and I may well be wrong on some.

A few smaller ones:

- `read-chain-state.md` Limitations mentions a direct WebSocket fallback for dev outside a Host. Does that exist? `chain-client.md` in this PR says there isn't one, and the package looks like it just throws.
- `store-data-on-chain.md` says `upload()` returns only the CID string. Isn't it wrapped in a `Result`?
- `{target=\_blank}` shows up on some new lines. AGENTS.md says to leave it to the MkDocs plugin, and a few are on internal links.
- 14 of the new pages are over the 160-char description target. `allowances.md` (183) and `register-dot-domain.md` (176) are the main ones.
- The contract sample appears in both `build/deploy-a-smart-contract.md` and `product-sdk/contracts.md`, and the two copies have drifted. Worth keeping one?
- 512B / 30s / ~2 weeks / ~60s are each repeated on 4-5 pages, mostly via one block copied into three build guides. Snippet include?
- The "today"/"yet" notes (AutoSigning, whoami, renewal handle) and the 10.4 PAS figure will go stale quietly. Fine to keep, just flagging.
- Not this PR, but `reference/apps/skills.md` links `product-sdk-bulletin`, which 404s. Looks like it's `product-sdk-cloud-storage` now.

Things I checked that all looked good: every internal link and anchor added here resolves, the 15 SDK API links and 11 RevX images are up, the three repos are public, the Claude Code plugin flow works exactly as written, appending `.md` really does return raw markdown, and the CLI surface (watchdog, `pg mod`/`logout`/`update`, aliases) matches the source.

One thing worth knowing: playground-cli deploy E2E has been red nightly for a couple of weeks, so the happy path isn't verified against live infra right now.

docs/apps/product-sdk/cloud-storage.md
```diff
@@ -0,0 +1,107 @@
...
+- To store bytes or files on chain and receive a CID plus a block receipt (`store(data).send()`).
+- To fetch content by CID inside a Host, with automatic reassembly of chunked uploads (`fetchBytes`, `fetchJson`).
+- To check an account's storage quota before uploading, so a missing allowance becomes an actionable error rather than a bare on-chain rejection (`checkAuthorization`).
+- Reads are Host-only; there is no public gateway fallback. To confirm a CID landed on chain without fetching the bytes, use `verifyStored`.
+
+## Core Concepts
+
+- **`CloudStorageClient`**: The main class. Create it with `CloudStorageClient.create({ environment, signer })`, then call `store`, `fetchBytes`, and `fetchJson`.
+- **Content addressing**: Uploads are identified by a CID (CIDv1) derived from their content, not by a path or key. The same bytes always produce the same CID.
+- **`store(data)` builder**: A fluent builder ending in `.send()`, which resolves with the CID, block number, and size. Large payloads are chunked automatically.
```
illegalcall: Worth noting which of these are optional? `size` looked like the only required field, and chunked stores (over ~2 MiB) come back with no `blockNumber` or `extrinsicIndex`, which is the renewal handle `deploy-your-app.md` sends people here to capture.

docs/apps/list-your-app.md
```diff
@@ -0,0 +1,121 @@
...
+
+```bash
+playground deploy --domain my-app --playground
+```
+
+With the phone signer, publishing adds one more approval in the Polkadot App, for writing the listing to the Playground registry.
+
+### Categorize the Listing
+
+Pass `--tag` to file your Product under a category in the directory. The value is one of `social`, `chat`, `defi`, `utility`, `gaming`, `marketplace`, or `irl`. Omit the flag and the CLI prompts you to pick one:
```
illegalcall: Is `defi` a valid tag? The CLI list looks like site/social/chat/utility/gaming/marketplace/irl, so `defi` may be rejected and `site` is missing here. Separately, is `my-app` the best example? A six-character base needs PoP Full, and quick-start pairs it with `--signer dev`.

docs/apps/concepts/networks.md
```diff
@@ -0,0 +1,53 @@
...
+# Networks
+
+## Introduction
+
+Two TestNet environments are available while you build Polkadot Products, and they are separate networks, not two names for one. The [Product SDK](/apps/product-sdk/) exposes both as presets:
+
+- **`paseo` (Paseo Next v2)**: The environment Polkadot Desktop development builds default to. It is a preview network and the successor to Paseo Next v1.
+- **`devnet`**: A third-party public Paseo TestNet, run by the [Polkadot Community Foundation](https://docs.polkadotcommunity.foundation/).
+
+Both expose the same core chains a Product uses — Asset Hub, the Bulletin Chain, and Individuality — so most Product code runs on either without changes. The production `polkadot` and `kusama` presets are not live yet; requesting them throws.
```
illegalcall: Is Individuality the same chain as the People Chain? The statement-store pages put allowances on the People Chain. If it's one chain under two names, could we gloss that here so readers can connect them?

docs/apps/index.md
```diff
@@ -62,11 +62,12 @@ Polkadot Apps is a complete environment for building Web3 decentralized applicat
 
 **Decentralized hosting**: Your bundle lives on a decentralized cloud storage provider. Your `.dot` name lives in a decentralized name service. There is no centralized hosting provider, no DNS service, no platform in the middle. Users fetch your Product directly and verify it themselves.
 
-**Product SDK**: The Product SDK covers what you would otherwise integrate yourself:
+**Product SDK**: The [Product SDK](/apps/product-sdk/) covers what you would otherwise integrate yourself:
 
 - **Chain access**: Query state and submit transactions through the Host. No RPC servers to operate; signing prompts open on the user's phone.
 - **Decentralized storage**: Upload files, get a permanent content-addressed URL.
 - **Real-time signed messaging**: Pub/sub between users of your Product via the [Statement Store](/reference/apps/infrastructure/statement-store/), every message verifiable.
+- **Smart contracts**: Deploy and call `pallet-revive` contracts on Asset Hub, resolved by name and typed from a manifest.
```
illegalcall: Does the SDK deploy contracts? `contracts.md` says deployment is cdm's job and the SDK only calls them. Reword to "call"?

docs/apps/quick-start.md
```diff
@@ -1,114 +1,156 @@
 ---
-title: Quick Start with RevX
-description: Generate and publish a Polkadot Product from the browser with RevX App Builder, then sign the final .dot deployment with the Polkadot App.
+title: Quick Start with the CLI
+description: Install playground-cli, pair it with the Polkadot App, build your Product, and deploy it to a live .dot name from the terminal.
 categories: Apps
 toggle:
   group: apps-quick-start
   canonical: true
```
illegalcall, line 7: Does moving `canonical: true` to the CLI variant change the published anchors? Looks like the RevX headings shift from `#open-revx` to `#revx-open-revx`, and fragments can't be redirected. Worth a quick check for external links?

docs/apps/build/deploy-a-smart-contract.md
bucanero: ✅ cdm has no phone signer — switched to `cdm deploy --suri` … / `cdm init`, with a warning that a missing `--suri` falls back to //Alice and first-writer-owns parks the name there.

docs/apps/build/deploy-a-smart-contract.md
bucanero: ✅ registry field is a no-op for deploy (resolved from -n) and 0xf62c… isn't canonical — dropped the "Set the Target Network" section entirely.

docs/apps/build/deploy-a-smart-contract.md
bucanero: ✅ `cdm deploy` doesn't write cdm.json, and shared-counter ships a populated contracts map — replaced the before/after JSON with an "Add the Contract to Your Manifest" section built around cdm install.

docs/apps/build/deploy-a-smart-contract.md
bucanero: ✅ Missing `ensureContractAccountMapped` and the query `.value` bug — the frontend sample now maps the account and checks .success before reading .value.

docs/apps/build/deploy-a-smart-contract.md
bucanero: Fixed ✅

docs/apps/get-started/set-up-your-ai-agent.md
bucanero: ✅ Added Solidity; it's supported via resolc/Foundry/Hardhat. Prompt now allows Rust or Solidity and bans mainnet-EVM tooling instead.

docs/apps/get-started/set-up-your-ai-agent.md
bucanero: ✅ Result shape — split the rule so agents check `.ok` for most calls but `.success` for contract query.

docs/apps/get-started/set-up-your-ai-agent.md
bucanero: ✅ Template files / `setup.sh` — scoped to the starter template and noted that moddable apps vary and many ship no setup.sh.

docs/apps/product-sdk/tx.md
bucanero: ✅ tx.md `result.value.ok` — confirmed unreachable (dispatch failures route to result.error as TxDispatchError); simplified to a single `result.ok` check.

docs/apps/product-sdk/index.md
bucanero: ✅ index.md umbrella — `statement-store` isn't re-exported and the import specifiers differ (`@parity/product-sdk/cloud-storage` vs `@parity/product-sdk-cloud-storage`); fixed both here and in build/index.md.

docs/apps/product-sdk/cloud-storage.md
bucanero: ✅ cloud-storage.md — noted `StoreResult` fields are optional and that chunked uploads return no blockNumber/extrinsicIndex (the renewal handle).

docs/apps/concepts/networks.md
bucanero: ✅  Added a gloss noting Individuality carries identity/personhood/Statement Store (what the reference docs call the People Chain).

docs/apps/index.md
bucanero: ✅ index.md (apps overview) "Deploy and call" → "call" (deployment is cdm's job).

docs/apps/quick-start.md
bucanero: Correct, the CLI-canonical shifts the RevX headings to #revx-…. I left CLI as the default since it's the forward path and the CLI anchors are new. 
RevX will be deprecated once the new Kiln tool is available, so anchors will change anyway.

docs/apps/list-your-app.md
bucanero: I stopped enumerating them since the tags might change in the future; the page now gives examples and says the CLI prompts with the current options.

bucanero: Also replaced `my-app` with `my-product` in quick-start and list-your-app

docs/apps/get-started/set-up-your-ai-agent.md
bucanero: ✅ Adjusted to the official Playground starter template. (not every app)

bucanero:
thanks for the review @illegalcall 👍 

I verified each point against the SDK/cdm source before changing anything. Nearly everything checked out. Fixes are pushed.

Btw, I also replied directly on your code comments and resolved each topic there.

**Smaller notes**
- ✅ `read-chain-state.md` "direct WebSocket fallback" — confirmed there's no such path (chain-client throws with no host provider). Corrected the Limitations line and the troubleshooting entry to say it throws; use the testing fakes out-of-Host.
- ✅ `store-data-on-chain.md` `upload()` — right, it's `Result`-wrapped. Reworded to "resolves with a `Result` whose value is only the CID string (no block or index)."
- ✅ `{target=\_blank}` — stripped across `docs/apps`, including the internal-link ones. Left to the plugin per AGENTS.md.
- ✅ Over-length descriptions — trimmed all 14 to ≤160 (`allowances` and `register-dot-domain` included).
- ✅ Contract sample drift — the build guide and `product-sdk/contracts.md` copies are now identical, and the build guide points to `contracts.md` as the canonical frontend reference.
- ✅ Repeated `512B / 30s / ~2 weeks / ~60s` block — extracted the storage-options block into a snippet (`.snippets/text/apps/storage-options.md`) included by the three build guides. (This also fixed a copy-paste bug where `store-data-on-chain.md` mismarked "Local KvStore (this page)".)
- ✅ `reference/apps/skills.md` `product-sdk-bulletin` 404 — fixed to `product-sdk-cloud-storage`.
- ℹ️ "today"/"yet" notes + 10.4 PAS — kept as intentional hedging, as you suggested.

ReinhardHatko (commented):
Three small things, verified against the published packages (`@parity/product-sdk-signer@0.14.2`, `@parity/result`) and against a local `mkdocs serve` build. Nice work on the contract-sample fixes — those all check out now.

docs/apps/product-sdk/signer.md
```diff
@@ -0,0 +1,102 @@
...
+
+```typescript
+import { SignerManager } from '@parity/product-sdk-signer';
+
+const manager = new SignerManager({ ss58Prefix: 0, dappName: 'my-product' });
+
+const connectResult = await manager.connect();
+if (!connectResult.ok) return; // HostUnavailableError outside a Host
+
+manager.selectAccount(connectResult.value[0].address);
```
ReinhardHatko: Two issues in this snippet, both of which bite on paste:

**`return` at module top level is a syntax error.** Line 41's `if (!connectResult.ok) return;` isn't inside a function, so this block doesn't compile as written.

**`connectResult.value[0]` can be `undefined` on a documented path.** `connect()` is `Promise<Result<SignerAccount[], SignerError>>`, and the SDK's own JSDoc calls out the empty case explicitly:

> If the host rejects the derivation (e.g. the identifier isn't registered), `connect()` resolves with an empty accounts list rather than throwing — consumers can still drive the explicit signing paths.

So an `ok` result with zero accounts is a real outcome, and `.value[0].address` throws there.

Wrapping the block fixes both:

```typescript
async function main() {
  const manager = new SignerManager({ ss58Prefix: 0, dappName: 'my-product' });

  const connectResult = await manager.connect();
  if (!connectResult.ok) return; // HostUnavailableError outside a Host

  const [account] = connectResult.value;
  if (!account) return; // host returned no accounts

  manager.selectAccount(account.address);

  const signature = await manager.signRaw(new TextEncoder().encode('hello'));
  if (signature.ok) {
    console.log(signature.value); // Uint8Array
  }
}
```

docs/apps/product-sdk/signer.md
```diff
@@ -0,0 +1,102 @@
...
+Use the `onConnect` hook to request resource allocations as soon as the connection is established, before any signing call:
+
+```typescript
+import { SignerManager } from '@parity/product-sdk-signer';
+
+const manager = new SignerManager({
+  ss58Prefix: 0,
+  dappName: 'my-product',
+  onConnect: async (_account, { requestResourceAllocation }) => {
+    await requestResourceAllocation([{ tag: 'AutoSigning', value: undefined }]);
```
ReinhardHatko: This matches the SDK's short JSDoc example verbatim, so it's faithful — but it omits two things the SDK's fuller example includes, and the page doesn't mention either.

**`requestResourceAllocation` throws.** Per `ConnectContext`, it "adapts `requestResourceAllocation` from `@parity/product-sdk-host` (which returns a `Result`) to this throwing contract — returns the unwrapped outcomes on success, throws the typed host error on failure." An unguarded `await` here can throw inside `onConnect`.

**The outcomes are discarded.** [`concepts/allowances.md`](https://github.com/polkadot-developers/polkadot-docs/blob/dapp-update/docs/apps/concepts/allowances.md) tells readers to "treat a non-`Allocated` outcome as a feature being unavailable rather than a fatal error" — but this example, the page's only one, gives them no pattern for doing that. The SDK's own richer example does both:

```typescript
onConnect: async (_account, { requestResourceAllocation, signal }) => {
  try {
    const outcomes = await requestResourceAllocation([
      { tag: 'AutoSigning', value: undefined },
    ]);
    if (signal.aborted) return;
    if (outcomes.some((o) => o !== 'Allocated')) {
      // degrade gracefully — a declined grant is not fatal
    }
  } catch (cause) {
    // typed host error
  }
},
```

Separately, and lower confidence since I can't check it from the type definitions: `allowances.md` says `AutoSigning` returns `NotAvailable` on both mobile wallets today and "do not build a flow that depends on it." If that's accurate, this page's one permission example demonstrates the one resource that never gets granted. A grantable resource might read better here.

redirects.json
```diff
@@ -1,5 +1,9 @@
 {
   "data": [
+    {
+      "key": "/apps/quick-start-cli/",
+      "value": "/apps/quick-start/"
+    },
```
ReinhardHatko, line 6: I think this redirect is a no-op — `/apps/quick-start-cli/` never served a 200 to begin with.

`awesome-nav` replaces `config.nav` with one generated from `.nav.yml` and clears `file.page` for every documentation page, so only files listed in a `.nav.yml` get a `Page` rebuilt — anything else is never rendered. `quick-start-cli.md` wasn't in `docs/apps/.nav.yml` on `master`, so it was never a reachable URL.

Confirmed against a local `mkdocs serve`: two unrelated pages absent from their `.nav.yml` (`smart-contracts/dev-environments/hardhat-polkadot.md` and `smart-contracts/cookbook/eth-dapps/uniswap-v2/core-v2-pvm.md`) both 404, and `/apps/quick-start-revx/` 404s today for the same reason — the toggle inlines the RevX variant into `/apps/quick-start/` rather than giving it its own URL, which I verified renders correctly with both variant buttons.

Harmless either way, so ignore if you'd rather keep it as a safety net for anything cached externally. Just flagging that it isn't catching a live URL.

docs/apps/product-sdk/index.md
```diff
@@ -0,0 +1,162 @@
...
+```typescript
+import { createApp } from '@parity/product-sdk';
+
+const app = await createApp({
+  name: 'my-app',
+  logLevel: 'info',
+});
+
+// Connect to host-provided accounts.
+const { accounts } = await app.wallet.connect();
```
ReinhardHatko: This snippet prints `Connected accounts: []` for almost anyone who runs it, because `name` is doing more than the comment suggests.

`createApp` passes `name` straight through as the signer's `dappName`:

```typescript
// packages/sdk/src/core/createApp.ts
const localKvStore = await createLocalKvStore({ prefix: config.name });
...
    dappName: config.name,
```

and per the umbrella CHANGELOG, that value is the dotNS identifier the host derives the product account from:

> The `dappName` you pass to `createApp({ name })` or `new SignerManager({ dappName })` is now also the dotNS identifier the host derives the product account from. `.dot` is appended automatically if missing. If your `dappName` isn't a valid registered dotNS identifier, the host will reject the derivation and `connect()` will resolve with `[]`.

`SignerManager.connect()`'s own JSDoc says the same — it "resolves with an empty accounts list rather than throwing." So with `name: 'my-app'`, the host tries to derive `my-app.dot`, fails unless the reader owns that name, and `connect()` succeeds with zero accounts. No error, just an empty log line.

The inline comment on line 47 ("namespaced under the app name in host storage") describes the *other* thing `name` does. `AppConfig.name`'s JSDoc covers both: "used to derive product accounts **and** namespace storage." The account-derivation half is the one with a failure mode, and it isn't mentioned.

Separately, `app.wallet.connect()` **throws** — it is not `Result`-typed like the rest of the page implies:

```typescript
// packages/sdk/src/core/createApp.ts — createWalletApi
async connect(): Promise<{ accounts: Account[] }> {
    const result = await signerManager.connect();
    if (!result.ok) {
        throw new Error(result.error.message);
    }
```

So outside a Host this line throws rather than returning something to check, and `new Error(result.error.message)` discards the typed error — a reader can't distinguish `HostUnavailableError` from a rejection.

Worth reconsidering what this example should be. As written, a reader's first encounter with the SDK is two calls that are the *only* two not following the documented `Result` convention, with a `name` that yields no account. Options: explain what `name` must be for derivation to succeed and guard both calls; or lead with the individual-packages path (`SignerManager`), which is `Result`-typed and matches what every Build guide teaches.

Minor, related: the prose here ("wires the wallet", line 32) and `useWallet` lead with `wallet`, which line 26 then explains is "the older name" for `signer`. Everything else in this section — the package, the guide, `SignerManager`, `getSigner()` — says signer. Reads a little odd to introduce the concept under the deprecated alias.

docs/apps/product-sdk/index.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Product SDK
+
+## Introduction
+
+The [Product SDK](https://github.com/paritytech/product-sdk) is the TypeScript SDK for building Polkadot Products. It gives your Product typed access to everything the platform provides: chain reads, transaction signing, decentralized storage, off-chain messaging, smart contracts, and identity, all routed through the Host your Product runs inside.
+
+The SDK never dials an RPC endpoint itself. Every sensitive operation (signing, chain access, storage) goes through the Host, which selects the network, holds the user's keys, and prompts for approval on the user's phone. Your Product calls a typed method; the Host mediates the rest.
+
+Fallible operations return a typed `Result` instead of throwing, so you check `.ok` before reading `.value`. This pattern runs through the whole SDK and is covered in each [Build guide](/apps/build/).
```
ReinhardHatko: > Fallible operations return a typed `Result` instead of throwing, so you check `.ok` before reading `.value`. This pattern runs through the whole SDK.

True of the individual packages, but not of the `createApp` surface this page demonstrates immediately below. Of the four sub-APIs on `App`, only one is `Result`-typed:

| `App` member | Convention |
|---|---|
| `wallet` | throws — `connect()` does `throw new Error(result.error.message)` |
| `localStorage` | no `Result`; `get` → `string \| null`, `set` → `void` |
| `chain` | throws — `getClient` is documented `@throws If the chain is not connected` |
| `cloudStorage` | **returns `Result`** (changed in the umbrella: "`createApp().cloudStorage.upload` / `fetch` now return `Result`") |

So the convention holds for `cloudStorage` and not the other three. Since this paragraph sits six lines above a snippet whose two calls are both from the non-`Result` group, a reader is primed to check `.ok` on values that don't have it.

Suggest scoping the claim — e.g. "Fallible operations in the individual packages return a typed `Result`… The `createApp` facade is thinner: `cloudStorage` returns a `Result`, while `wallet`, `localStorage`, and `chain` throw." Whatever the wording, worth not stating it as universal, because the one surface where it's mostly false is the one this page opens with.

ReinhardHatko (commented):
Two more on `product-sdk/index.md`, both about the `createApp` surface. Verified against `paritytech/product-sdk` @ `main`, the published `@parity/product-sdk-signer@0.14.2`, and the umbrella package's CHANGELOG. The umbrella-subpath section added in the last round is a real improvement — the `wallet`→`signer` note in particular.

docs/apps/product-sdk/index.md
```diff
@@ -0,0 +1,162 @@
...
+- **`useProductSDK`**: The `App` instance and connection state.
+- **`useWallet`**: The connected account and signing helpers.
+- **`useLocalStorage`**: Reactive per-Product key-value storage.
+- **`useChain`**: The host-routed chain client.
+
+The [Shared Todo App tutorial](/apps/tutorials/shared-todo-app/) uses these bindings end to end.
+
+## Testing Without a Host
+
+Because `createApp` and the host-only methods require a Host, the SDK ships fakes for local development and automated tests. The `@parity/product-sdk/testing` subpath provides `createFakeApp` and per-capability fakes (`createFakeSignerProvider`, `createFakeHostLocalStorage`, and more), so you can exercise Product logic in a plain Node or browser test without Polkadot Desktop.
```
ReinhardHatko: Four things here. Checked `packages/sdk/src/testing.ts` at `main` and the published `/testing` bundles.

**1. There is no chain fake, and `app.chain` throws.** `createFakeApp` fakes `wallet`, `localStorage`, and `cloudStorage` only. Every `chain` method fails:

```javascript
"createFakeApp: `chain` is not configured. Pass a `chain` override in options. " +
"There is no chain-read fake — the host owns RPC selection; test chain-reading " +
"logic behind your own interface and cover the real wiring in e2e."
```

So "exercise Product logic ... without Polkadot Desktop" doesn't hold for chain reads, which is the first Build recipe. The SDK also states the intended pattern — put chain reads behind your own interface, cover the real wiring in e2e. Worth passing that on.

**2. "local development" overstates it.** The module header scopes these to tests:

> Test fakes for `@parity/product-sdk` (umbrella). `createFakeApp` returns a fake `App` — use it directly in logic tests or hand it to `ProductSDKContext.Provider` for React component tests.

For developing in a browser, the answer is already in this section: run the Product inside Desktop from `localhost` per [Set Up Your Project](/apps/build/#set-up-your-project). Suggest dropping "local development" so the fakes read as a test tool.

**3. `statement-store` fakes are not re-exported.** Deliberate, per a comment in `testing.ts`: re-exporting would add a dependency the umbrella doesn't have and could skew against the version the app installs. Import `@parity/product-sdk-statement-store/testing` directly. This page already flags that statement-store isn't in the umbrella; same caveat applies here.

**4. `createFakeApp` is synchronous**, while real `createApp` is `async`. Harmless, but the two aren't drop-in symmetric.

No problem with the browser claim itself — I checked for Node-only imports in the `/testing` bundles and found none (`contracts/testing.js` pulls only `viem`). `createFakeSignerProvider` and `createFakeHostLocalStorage` both exist as named.

ReinhardHatko (commented):
One more, on the testing section. The signer.md and sign-and-submit.md fixes in 018026b0 look good — the `AutoSigning` callout and the `ok([])` warning both go further than what I flagged.

docs/apps/product-sdk/index.md
```diff
@@ -0,0 +1,211 @@
...
+The subpath also re-exports the per-package fakes for [`signer`](/apps/product-sdk/signer/), [`local-storage`](/apps/product-sdk/local-storage/), [`contracts`](/apps/product-sdk/contracts/), and [`host`](/apps/product-sdk/host/) — `createFakeSignerProvider`, `createFakeHostLocalStorage`, and the rest — so one import covers them.
+
+!!! note "Statement store fakes are imported separately"
+    They are deliberately not re-exported here, for the same reason `statement-store` has no umbrella subpath: adding it would pull in a dependency the umbrella does not otherwise have, and could pin a different version than the one your Product installs. Import them from `@parity/product-sdk-statement-store/testing` instead.
+
+Individual packages also expose a dev path where it makes sense; for example, `SignerManager.connect('dev')` loads the standard Substrate dev accounts. See [Sign and Submit Transactions](/apps/build/sign-and-submit/#test-without-a-host).
+
+## Requirements
+
+- **Node.js**: version 20 or later.
```
ReinhardHatko: This is wrong for two of the packages now documented in this section.

`terminal.md:101` requires **Node 21**, and `auth.md` inherits it ("subject to the same Node 21 requirement through it"). Both are correct — the terminal package's README says so:

> **Requires Node ≥21.** The package relies on the global `WebSocket` exposed by Node 21+ (via `@polkadot-api/ws-provider@0.9`). On older Node versions the WebSocket connection fails at runtime with `WebSocket is not defined`.

So a CLI author reads "Node.js: version 20 or later" here, believes they are fine, and fails at connect time.

Suggest noting the exception, e.g.:

> - **Node.js**: version 20 or later, except [`terminal`](/apps/product-sdk/terminal/) and [`auth`](/apps/product-sdk/auth/), which require 21 or later.

docs/apps/product-sdk/terminal.md
```diff
@@ -0,0 +1,136 @@
...
+    throw error;
+  }
+}
+```
+
+Both helpers default `sessionId` to the only paired session. With zero or more than one paired session and no explicit id, they throw `AllowanceError` with `reason: 'NoSession'`.
+
+## Limitations
+
+- **Requires Node 21 or later.** The package relies on the global `WebSocket` that Node 21 exposes; on older versions the connection fails at runtime with `WebSocket is not defined`.
```
ReinhardHatko: Accurate, but the package README offers a way out that this drops:

> Older Node versions (18, 20) will fail at connect time with `WebSocket is not defined` — upgrade Node, or pass an explicit `websocketClass` from the [`ws`](https://www.npmjs.com/package/ws) package.

Worth adding the `websocketClass` option — someone pinned to Node 20 by other constraints currently reads this as a hard block.

docs/apps/product-sdk/auth.md
```diff
@@ -0,0 +1,156 @@
...
+  await auth.waitForLogin(connection.login, (status) => {
+    if (status.step === 'pending') console.log(status.stage);
+  });
+}
+
+const session = await auth.getSessionSigner();
+if (session) {
+  console.log('signing as', session.addresses.productAddress);
+  // Use session.signer with submitAndWatch, a contract .tx(), or any PAPI call.
+  session.destroy(); // required — releases the WebSocket
```
ReinhardHatko: Since the page makes the point twice that `destroy()` is mandatory or the CLI hangs, this example should model it. `destroy()` here only runs if `session` is truthy and nothing above threw — whereas the `resolveSigner` example further down correctly uses `finally`.

```typescript
const session = await auth.getSessionSigner();
if (session) {
  try {
    console.log('signing as', session.addresses.productAddress);
    // Use session.signer with submitAndWatch, a contract .tx(), or any PAPI call.
  } finally {
    session.destroy(); // required — releases the WebSocket
  }
}
```

ReinhardHatko (commented):
Reviewed the three new pages (`auth.md`, `individuality.md`, `terminal.md`) against the published packages — `auth@0.2.8`, `terminal@0.8.0`, `individuality@0.2.0`. Everything load-bearing checks out: `DEFAULT_RESOURCES` is exactly `BulletInAllowance` / `StatementStoreAllowance` / `SmartContractAllowance` with index 0, `AuthConfig` has precisely those four fields, `resolveSigner` does take the client first, all seven personhood state tags match, and all three `AsPersonInfo` variants are right. The tagged-object-vs-bare-string warning in `auth.md` is correct — `summarizeOutcomes` compares `outcome.tag` and buckets by `resources[i]`, so "order-sensitive" holds too. Also noticed `api/individuality` 404s on the API site and the page links to the repo instead, unlike the other two — nice catch by whoever wrote it.

Three small things below.

docs/apps/build/sign-and-submit.md
```diff
@@ -190,42 +230,53 @@ import {
...
 
-const selectResult = manager.selectAccount(result.value[0].address);
-if (!selectResult.ok) return;
+  const signResult = await manager.signRaw(new TextEncoder().encode('test'));
+  if (!signResult.ok) return null;
 
-const signResult = await manager.signRaw(new TextEncoder().encode('test'));
-if (signResult.ok) {
-  console.log(signResult.value); // Uint8Array - the raw signature
+  return signResult.value; // Uint8Array - the raw signature
```
ReinhardHatko, line 279: The same fact is stated twice, five lines apart, and both copies are incomplete.

Line 284, the warning closing this section:

> `getProductAccount`, `getProductAccountAlias`, and `createRingVRFProof` are host-only APIs. They return `HostUnavailableError` when the active provider is `'dev'`.

Line 288, the first Limitations bullet:

> `getProductAccount`, `getProductAccountAlias`, and `createRingVRFProof` require an active host connection. They are not available in `'dev'` mode.

Same three methods, same constraint. The warning is the better of the two — it names `HostUnavailableError`, where the bullet only says "not available."

Both also omit a fourth method. `product-sdk/signer.md:99` has the full set:

> `getProductAccount`, `getProductAccountAlias`, `createRingVRFProof`, and `getUserId` are Host-only.

`getUserId` matters here specifically: [`concepts/identity.md`](/apps/concepts/identity/) recommends it as the way to read a personhood username for a display name, so a reader following that advice under the dev provider hits an error this page implied would not occur.

Suggest dropping the Limitations bullet, keeping the warning, and adding `getUserId` to it.

ReinhardHatko (commented):
One duplication just below this snippet. Anchored on the nearest diff line — the lines in question (284, 288) are outside the diff, so GitHub will not take a comment directly on them.

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,276 @@
...
+
+const manager = ContractManager.fromClient(
+  cdmJson,
+  client.raw.assetHub,
+  paseo_asset_hub,
+  { signerManager },
+);
+
+// pallet-revive requires each signing account to be mapped once.
+await ensureContractAccountMapped(manager.getRuntime(), account.address, signer);
```
ReinhardHatko: This snippet does not compile as written. `signerManager` (line 216), `account` and `signer` (this line), and `signer` again on line 231 are never defined.

The prose below covers where they come from — "the same `SignerManager` you set up for signing, and `account` and `signer` come from it" — but a reader pasting the block gets three unresolved identifiers.

Worth matching what `signer.md` and `sign-and-submit.md` got in `018026b0`: wrap it in an `async function` and derive the three values, so the block stands alone like the rest of the section now does.

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,276 @@
...
+
+!!! note "query returns a status, tx returns a Result"
+    `query` is a dry run that does not throw on a revert; branch on `.success`, because on failure `.value` holds the dispatch-error payload, not your data. `tx` returns a `Result`; check `.ok` before assuming the write landed.
+
+## Redeploy After a Contract Change
+
+Contracts are immutable once deployed. Changing contract code means deploying a new version: run `cdm deploy -n paseo --suri ...` again (or choose `yes` at the contract prompt in `playground deploy`). The registry appends a new version under the same name; run `cdm install @my-app/counter -n paseo` afterward to refresh the address and ABI in `cdm.json` so your frontend picks up the new deployment.
+
+Because a redeploy gives your contract a new on-chain address, existing users pointed at the old address keep using the old contract until they load the new bundle. For a contract with live users and stored state, plan the migration deliberately rather than redeploying in place.
```
ReinhardHatko: > For a contract with live users and stored state, plan the migration deliberately rather than redeploying in place.

This is the only mention of migration on the page, and nothing explains what planning one involves — no steps, no link. Telling a reader to plan something the docs do not cover leaves them stuck at exactly the point the warning is meant to help.

Either drop the sentence, or give it somewhere to go. The surrounding paragraph already supplies the mechanics that matter (new address, old users on the old contract until they reload), so a short list — snapshot state before the redeploy, read from the old address until the new one is populated, keep the old name resolvable — would close it without much added length.

docs/apps/build/deploy-a-smart-contract.md
```diff
@@ -0,0 +1,276 @@
+---
+title: Deploy and Integrate a Smart Contract
+description: Add a PolkaVM smart contract to your Polkadot Product — deploy it to Asset Hub with the Contract Dependency Manager and call it from your frontend.
+categories: Apps
+page_badges:
+  tutorial_badge: Advanced
+---
+
+# Deploy and Integrate a Smart Contract
```
ReinhardHatko: Scope comment on the page as a whole, not a defect.

The title promises deploy and integrate, but most of the page is `cdm` tooling. By section length:

| | Lines |
|---|---|
| `cdm` tooling — Install cdm, Set Up a Signing Account, Verify the Toolchain, Build and Deploy, Add to Manifest, Redeploy | 88 |
| Contract work — Scaffold, Call from Frontend | 110 |

Closer than it reads, but three of those `cdm` sections (Install, Signing Account, Verify Toolchain) are new since the first draft, so the balance moved toward tooling as the page was corrected. A reader arriving from the Build index — where every other recipe is one SDK capability — meets a toolchain install, a keypair generation step, and a network-preset warning before reaching any contract code.

No objection to the content; the `cdm` material is accurate and needed. But it might belong in its own page, with this one reduced to writing a contract, deploying it, and calling it — mirroring how `product-sdk/*` pages stay one-capability. Worth a deliberate decision either way, since the title currently undersells what the page asks of you.

ReinhardHatko (commented):
Three on `deploy-a-smart-contract.md`. The earlier round fixed a lot here — Rust-or-Solidity, the `--suri` correction, `cdm install` writing the manifest, and the `.success` check all read correctly now.

UtkarshBhardwaj007 (approved):

PR #1753, opened 2026-09-09
Update dotNS docs for contracts v0.8.0 and CLI v0.9.2

### LINKED ISSUE #1751: [Docs] dotNS pricing, personhood and namehash pages describe incorrect/outdated behavior
### Page URL

https://docs.polkadot.com/reference/apps/infrastructure/dotns/poprules-pricing/

### What’s the issue?

Also affected, same root cause:

- https://docs.polkadot.com/reference/apps/infrastructure/dotns/
- https://docs.polkadot.com/reference/apps/infrastructure/dotns/name-mechanism/
- https://docs.polkadot.com/reference/apps/infrastructure/dotns/architecture/
- https://docs.polkadot.com/apps/concepts/identity/

The dotNS team has reviewed these pages against the code, the pricing page describes a cost model that has never been deployed, and several statements elsewhere are outdated or incorrect. Grouped by severity.

### 1. Deposits do not vary with length

The deployment script registers that model with the base deposit constant, currently 10 DOT, and the deposit is refundable. However, the contracts allow governance upgrades to the economics model and more complex pricing models will probably be deployed in future releases.
Anyway, the two formulas on the page, `startingPrice × (15 − nameLength)` and `startingPrice / 2`, are not implemented anywhere.

### 2. The bands above eight characters are wrong

Classification has three bands, not six, and no boundary at 14 or 15 characters:

- base length 5 or fewer: governance-reserved, rejected on the public path
- base length 6 to 8: requires personhood
- base length 9 or more: open to anyone

### 3. A lite personhood name now keeps its separator: `alice.01`, not `alice01`

This is a recent change and the page predates it. A lite person username issued by the personhood gateway is stored, minted and hashed as `stem.NN`: a stem of lowercase ASCII letters, one separator, then exactly two digits. It is one whole label, not a dotted path.

Two consequences for the table:

- "6-8 chars + 2-digit suffix (e.g. `alice01`)" is no longer a lite shape. Digits are no longer stripped from an ordinary label.
- `hamilton01` is open to anyone because it is 10 characters, not because it carries a suffix. The suffix shape no longer affects classification for names bought on the public path.

Lite names are also not purchasable: the public registration path rejects any label containing a separator, so a lite name can only be issued through the personhood gateway.

### 4. "Self-declared PoP tier" is not what the contract does

The admonition on this page and on the dotNS index page says dotNS reads the PoP tier "from a status the user sets themselves", with on-chain verification "forthcoming". However, there is no user-settable status anywhere in the contracts, and there never has been.

### 5. Namehash: the ENS rule alone gives the wrong key for a lite name

`name-mechanism/` documents plain ENS recursive namehash. That is correct for ordinary names, but for a lite personhood name the separator is *inside the label*: `michael.42` is hashed as a single label under the TLD.

### 6. Not every name is transferable

`apps/concepts/identity/` says ownership "is held by an Asset Hub account and is transferable". That holds for public registrations only. Every name issued by the personhood gateway is now soulbound and permanently non-transferable.

### 7. Architecture page: renewals do not exist

The page is marked provisional, so this is low priority, but two items are incorrect. There is no renewal contract or sub-mechanism: registrations are permanent by construction, with no expire, renew or release path on the registrar. And there is no separate transfer contract; transfers are handled on the registrar itself through its fee-on-transfer hook and the escrow, and soulbound names cannot transfer at all.

### Suggested improvement

- Replace the pricing ladder with the latest model, also stating this may be upgraded by governance
- In `name-mechanism/`, add a carve-out under Layer 2 stating that a lite personhood name is hashed as one whole label
- In `apps/concepts/identity/`, qualify transferability according to the latest releases
- Address remaining issues

Happy to review a draft of the corrected pages, and to supply the exact constants and band boundaries if it helps.

re-gius:
## 📝 Description

Brings the dotNS documentation in line with dotNS contracts `v0.8.0` and `@parity/dotns-cli` `0.9.2`. Closes #1751.

The pages had drifted from the contracts on several points at once, and the issue collected them. The largest is pricing: the reference described a scarcity ladder with free tiers and two deposit formulas, and no deployed cost model has ever worked that way. The rest follows from three behaviour changes in the release: a personhood username now keeps its separator and is hashed as one whole label, gateway-issued names are soulbound, and a name's length is counted as written rather than with a digit suffix stripped off. Registrations are permanent, so the renewal contract and renewal commands that appeared across several pages do not exist.

## 🔍 Review Preference

Choose one:
- [x] ✅ I have time to handle formatting/style feedback myself 
- [ ] ⚡ Docs team handles formatting (check "Allow edits from maintainers")  

## ✅ Checklist

- [x] Changes tested  
- [x] [PaperMoon Style Guide](https://github.com/papermoonio/documentation-style-guide) followed

### DIFF
--- a/docs/reference/apps/infrastructure/dotns/index.md
+++ b/docs/reference/apps/infrastructure/dotns/index.md
@@ -16,15 +16,22 @@ Four properties shape how a Product developer interacts with dotNS:
 
 - **The registry lives on Asset Hub**: Names, owners, and the content references they point at are stored as contract state on Asset Hub, not on the People Chain or Bulletin Chain.
 - **Name resolution is content-addressed at the end**: A `.dot` name resolves to a CID, and the CID points at bytes on the [Bulletin Chain](/reference/apps/infrastructure/bulletin-chain/) (or via an IPFS gateway). See [Name Mechanism](/reference/apps/infrastructure/dotns/name-mechanism/).
-- **Pricing is personhood-gated by PopRules**: Short names are free for personhood holders; longer or numerically-suffixed names are open to anyone but require a deposit. See [PopRules and Pricing](/reference/apps/infrastructure/dotns/poprules-pricing/).
+- **Eligibility is personhood-gated by PopRules**: A name's length places it in a band. Six to eight characters requires PoP Full on the public path, and five or fewer is not sold there at all. Every name the public path admits pays the same refundable deposit, so personhood gates who may register rather than what it costs. See [PopRules and Pricing](/reference/apps/infrastructure/dotns/poprules-pricing/).
 - **The architecture is a small set of cooperating contracts**: Not a single registrar — a set of contracts each handling a slice of the model. See [Architecture](/reference/apps/infrastructure/dotns/architecture/).
 
 For the Product-side how-to (registering a name, publishing your bundle), see [Register and Publish](/apps/deploy-your-app/).
 
-!!! warning "Known gaps"
-    Two operational caveats apply to the current dotNS surface:
+## One Deployment per Network
 
-    - **Self-declared PoP tier**: dotNS reads PoP tier from a status the user sets themselves. On-chain verification of PoP tier against the People Chain is a forthcoming integration; until then, treat the tier check as cooperative, not adversarial.
+dotNS is deployed separately on each network, and three things follow from that.
+
+- **The TLD is set at deploy time**: Polkadot uses `.dot`, and a TestNet uses its own, so the same name is `awesome.dot` on Polkadot and `awesome.paseo` on Paseo. Tools take the bare label and append the TLD of the environment they target, so a label carrying a different TLD is rejected rather than translated.
+- **The deposit is the network's native token**: The amount is the same on every network, and the token is whatever that chain uses, so a registration costs 10 DOT on Polkadot and 10 PAS on Paseo. See [PopRules and Pricing](/reference/apps/infrastructure/dotns/poprules-pricing/).
+- **Registrations do not cross networks**: Each deployment keeps its own registry, so holding a name on a TestNet grants no claim to the matching name on Polkadot. A name has to be registered on each network where it is needed.
+
+!!! warning "Known gaps"
+    One operational caveat applies to the dotNS surface:
+
     - **Operational runbook gaps**: Some operator-side procedures (migration, batch updates, contract upgrade paths) are not yet documented for this build. The reference here covers the developer-facing surface.
 
 ## Where to Go Next
@@ -51,7 +58,7 @@ For the Product-side how-to (registering a name, publishing your bundle), see [R
 
     ---
 
-    The pricing ladder for name registration: who can register what at what cost, organized by name length and PoP tier.
+    Which length band a name falls in, who may register it, and what the deposit is.
 
     [:octicons-arrow-right-24: Reference](/reference/apps/infrastructure/dotns/poprules-pricing/)
 
--- a/docs/reference/apps/infrastructure/dotns/index.md
+++ b/docs/reference/apps/infrastructure/dotns/index.md
@@ -1,6 +1,6 @@
 ---
 title: dotNS Reference
-description: Reference for dotNS — the .dot name system on Asset Hub that resolves Product names to published bundles, with PopRules pricing and contract architecture.
+description: Reference for dotNS — the .dot name system on Asset Hub that resolves Product names to published bundles, with PopRules eligibility and contract architecture.
 categories: Apps, Reference
 ---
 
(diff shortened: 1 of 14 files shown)


bucanero:
@re-gius thanks a lot for the PR 🙏 

I have some questions for the dotNS team so I can adjust and clarify a few points over the docs:

1. **`isPopIssued(label)`** — is the argument the label string, or the namehash/node? `name-mechanism.md` exists precisely to explain that labels and nodes differ, so this signature needs to be exact.
2. **Is the short-name market open on Paseo today?** The docs currently read both ways , and the answer changes what `troubleshooting.md` should tell users.
3. **Is 10 DOT the real MainNet deposit, and is dotNS deployed on Polkadot at all yet?** `index.md` states "a registration costs 10 DOT on Polkadot" as present fact; if it's TestNet-only, that should be framed as design.
4. **Can a stem of five or fewer get a Lite username?** `poprules-pricing.md` implies `alice.42` is refused by the gateway — a surprising constraint worth confirming and stating explicitly.
5. **Does the reserved/governance path have a three-character floor?** The claim "Nothing shorter than three characters can be registered at all" is broader than the two sentences supporting it, which only cover the public path and the gateway.
6. **Which CLI enforces label rules under `pg deploy`, and from what version?** 
7. **Escrow lifecycle**: definitions for release / redeem / reclaim / withdraw, and who receives a refunded deposit after a transfer (`transfer.md:28` says the deposit follows the name).


Also, a few assumptions to do a sanity-check:

- That the public path is "the commit-reveal route the CLI and the deploy flow use."
- That most transfers are free — deduced from "two independent conditions each trigger the charge." If the real default is charged-unless-exempt, `transfer.md:32` is now wrong in the opposite direction.
- That the lens contract lists "the names issued to one person," and that the registrar is itself the ERC-721 carrying ownership — both read off the address table into `architecture.md`.

re-gius:
> @re-gius thanks a lot for the PR 🙏
> 
> I have some questions for the dotNS team so I can adjust and clarify a few points over the docs:

Thanks for the careful pass @bucanero. Answers below, each checked against the contracts source at the latest release v0.8.0. One note first: the branch has been updated since your review. The docs now say "device name" and "personhood name" for the two gateway-issued kinds, with the tiers Lite and Full defined once in PopRules and Pricing, and several of the answers below are already folded into the pages.

>     1. **`isPopIssued(label)`** — is the argument the label string, or the namehash/node? `name-mechanism.md` exists precisely to explain that labels and nodes differ, so this signature needs to be exact.

`isPopIssued` takes the label string, not the node, so for a device name you pass the dotted form, for example `joseph.42`. The name-mechanism page now states this next to the hashing rule.

>     2. **Is the short-name market open on Paseo today?** The docs currently read both ways , and the answer changes what `troubleshooting.md` should tell users.

`shortNamesEnabled` is a Root-gated operations knob, default `false`, that testnet operators use to seed reservations. It is not a market a user can wait for, so we removed every mention of it from the docs. The user-facing rule is simply that below nine characters nothing is sold on the public path, and six to eight characters needs full proof of personhood.


>     3. **Is 10 DOT the real MainNet deposit, and is dotNS deployed on Polkadot at all yet?** `index.md` states "a registration costs 10 DOT on Polkadot" as present fact; if it's TestNet-only, that should be framed as design.

dotNS runs on TestNets only today. The deposit on Paseo is 10 PAS; 10 DOT is the designed value for a future Polkadot deployment. index.md has been reframed to say exactly that.

>     4. **Can a stem of five or fewer get a Lite username?** `poprules-pricing.md` implies `alice.42` is refused by the gateway — a surprising constraint worth confirming and stating explicitly.

No: a stem of five or fewer cannot get a device name. The gateway classifies by the stem, meaning the label with its allocated two-digit suffix removed, and a stem of five or fewer lands in the governance-reserved band, so `alice.42` is refused (`InvalidLiteLabel` in `DotnsPopController`). `poprules-pricing.md` now states the stem rule explicitly.

>     5. **Does the reserved/governance path have a three-character floor?** The claim "Nothing shorter than three characters can be registered at all" is broader than the two sentences supporting it, which only cover the public path and the gateway.

The three-character floor is a public-path check, enforced in `DotnsRegistrarController` (`LabelTooShort`). The gateway pallet applies no floor of its own; its stems bottom out at six because five and fewer classify as reserved. The governance path is Root-gated and not bound by the controller check. The page has been rescoped accordingly and no longer makes the broad claim.

>     6. **Which CLI enforces label rules under `pg deploy`, and from what version?**

The chain is always authoritative. `PopRules` accepts a lowercase ASCII label or a device-name shape, with no cap on trailing digits since contracts v0.7.0. Client-side, @parity/dotns-cli carried the old two-digit cap and lifted it in 0.9.0. I will confirm which client-side validation `pg deploy` itself bundles and follow up here.

>     7. **Escrow lifecycle**: definitions for release / redeem / reclaim / withdraw, and who receives a refunded deposit after a transfer (`transfer.md:28` says the deposit follows the name).

Escrow lifecycle, all holder-driven except the last step:

- `release(tokenId)`: the current holder moves the name into escrow custody. It stamps two clocks, a withdrawal cooldown and a redeem window.
- `redeem(tokenId)`: the undo. During the redeem window, and only while the deposit has not been withdrawn, the releasing holder takes the name back. No value moves.
- `withdraw(tokenId)`: after the cooldown, credits the refundable deposit to the holder's pull-payment balance; `claimWithdrawal()` pulls it.
- `reclaim(tokenId, newOwner)`: controller-only, once the redeem window has elapsed. It hands the name to a new registrant and settles any outstanding deposit to the previous holder's balance, so an absent holder keeps the value while the name recycles.
- After a transfer the deposit follows the name: the escrow rebinds the refund recipient to the new holder on every transfer and refunds nothing at transfer time, so `transfer.md` line 28 is correct.

> Also, a few assumptions to do a sanity-check:
> 
>     * That the public path is "the commit-reveal route the CLI and the deploy flow use."
> 
>     * That most transfers are free — deduced from "two independent conditions each trigger the charge." If the real default is charged-unless-exempt, `transfer.md:32` is now wrong in the opposite direction.
> 
>     * That the lens contract lists "the names issued to one person," and that the registrar is itself the ERC-721 carrying ownership — both read off the address table into `architecture.md`.

yes, the public path is the commit-reveal route in `DotnsRegistrarController`, which the CLI and the deploy flow drive. Yes, the default transfer is free; the floor is zero unless the recipient fails the name's band requirement or the recipient's personhood tier is lower than the sender's, exactly as `transfer.md` now describes. And both architecture reads are right: DotnsPopLens lists the gateway-issued names of one account (liteNamesOf / fullNamesOf, each entry verified against ownership and isPopIssued), and DotnsRegistrar is itself the ERC-721 carrying ownership.

bucanero (approved):

docs/reference/apps/infrastructure/dotns/index.md
```diff
@@ -10,22 +10,29 @@ categories: Apps, Reference
...
+
+- **The TLD is set at deploy time**: Each network's registry carries one TLD. A TestNet uses its own, so a name reads `awesome.paseo` on Paseo, and the planned Polkadot deployment uses `.dot`. Tools take the bare label and append the TLD of the environment they target, so a label carrying a different TLD is rejected rather than translated.
+- **The deposit is the network's native token**: A registration costs 10 units of whatever the chain uses, so 10 PAS on Paseo. See [PopRules and Pricing](/reference/apps/infrastructure/dotns/poprules-pricing/).
+- **Registrations do not cross networks**: Each deployment keeps its own registry, so holding a name on a TestNet grants no claim to the matching name on Polkadot. A name has to be registered on each network where it is needed.
+
+The deployments have different operators. Parity runs dotNS on the TestNets, including Paseo, and the planned deployment on Polkadot will be run by the [Polkadot Community Foundation](https://docs.polkadotcommunity.foundation/). The contracts are the same, but each operator configures its own deployment, so governance-set values such as the deposit amount need not match between them.
```
re-gius: The deposit is a weak example here, because `DotnsFlatPricing.deposit` is immutable and fixed in the constructor; governance changes it by repointing `DotnsCostModelRegistry` at a different contract.
So I would change this to **owner/deployer-set values**

docs/reference/apps/infrastructure/dotns/escrow.md
```diff
@@ -0,0 +1,78 @@
...
+
+`release(tokenId)` moves the name into escrow custody. It starts two independent clocks: a withdrawal cooldown, which gates when the deposit can be taken back, and a redeem window, which gates how long the holder can change their mind.
+
+Releasing is not a sale, and it costs nothing beyond gas. A transfer into or out of escrow is exempt from the transfer fee. See [Name Transfers](/reference/apps/infrastructure/dotns/transfer/).
+
+### Redeem
+
+`redeem(tokenId)` is the undo. While the redeem window is open, and only while the deposit has not been withdrawn, the releasing holder takes the name back. No value moves, because the deposit never left escrow.
+
+Withdrawing forecloses this. Once `withdraw` has run, the name cannot be redeemed even if the redeem window is still open.
```
re-gius:
```suggestion
Withdrawing a funded deposit forecloses this. Once `withdraw` has paid one out, the name cannot be redeemed even if the redeem window is still open.
```

Mentioning funded deposits because this does not apply to zero-deposit names, which are an exception so it's not necessary to mention them, but the new formulation makes the sentence correct.

re-gius (commented):
LGTM - just two final comments

PR #1754, opened 2026-09-17
dotNS update

bucanero:
## Apps docs: identity and registration reframes, dotNS normalization

Follow-up to #1740, incorporating reviewer feedback on the Apps section. Merged with master, so it includes the dotNS v0.8.0 / CLI 0.9.2 update (#1753) and builds on it rather than around it.

### Register a `.dot` Domain — tool-neutral

The page read as "how `playground deploy` registers a name," though registration is a chain-level operation and three CLIs perform it. Adds a **Ways to Register** comparison plus a section each for `playground`, `pad` (`@parity/polkadot-app-deploy`), and the dotNS CLI, and promotes the dotNS CLI out of its "provisional, last resort" framing into the owner of everything past registration.

Master's name rules, personhood tiers, and deposit model are carried through unchanged — the reframe is structural, not factual.

### Identity — two user identities, not three

The page listed the Product's `.dot` name as one of a user's three identities, then contradicted itself twice ("it is not how your Product identifies a user"). Now leads with the two identities a user actually has — the per-app account and Proof of Personhood — and covers the Product's `.dot` address separately as the thing that is explicitly *not* one. Adds a **The Earned Name** subsection using master's terminology, and moves the display-name how-to into `sign-and-submit.md` as **Show a Display Name**, where the `getUserId` host-only and permission-prompt caveats now sit alongside the recommendation.

### Consistency

- `DotNS` → `dotNS` across the docs, matching the reference page's own H1. Skips derivation identifiers (`playground.dot/0`), contract names (`DotnsPopController`), and URLs.
- The `.paseo` TLD warning now appears on all five deploy-path pages, including RevX; snippet reworded so it reads correctly on a page with no CLI.
- List Your App notes that the Playground directory is scoped to the TestNet the CLI targets, and that Browse is the destination for a Product ready for end users.
- `.vale.ini`: disables built-in `Vale.Terms`. It enforced vocabulary casing with no notion of headings, sentence position, proper nouns, or domain names — 608 errors across `docs/`, effectively all false positives (`## Runtime`, `polkadot.com`, `polkadot-sdk`). House terminology is still enforced by `PaperMoon.Terminology`. This makes 73 genuine errors visible that the noise was burying.


bucanero:
@re-gius could you take a look? 🙏  thank you!

docs/apps/concepts/identity.md
```diff
@@ -42,33 +35,51 @@ Your Product obtains this account through the [`signer`](/apps/product-sdk/signe
...
 
 - **A tier**: The strength of the proof, from none, through an attested proof of a unique device, to full personhood, proven in the Polkadot App. The [Proof of Personhood reference](/reference/apps/infrastructure/pop/) names the tiers.
 - **A per-app alias**: A Ring-VRF-derived identifier that is deterministic for a given user and Product, and unlinkable across Products.
+- **An earned name**: The human-readable name the personhood gateway issues to a user who has proved a tier, such as `joseph.42`.
 
 An alias is never an account address, and — like the per-app account — it is scoped per Product so it cannot be used to correlate a user across Products. Cross-Product alias linking requires an explicit consent step. Use personhood to gate features on verified-human status (for example, one action per person) without learning who the user is.
 
 !!! warning "Names and usernames can be coupled"
     The three identities are separate by design, but not always independent in practice: the personhood gateway issues a device name into dotNS naming, which couples that username to a dotNS name of the same text. Do not assume the identities can never be linked.
```
re-gius: The reframe left this behind: the page now opens "carries two identities" (lines 3 and 11), but this still says three. 

Also worth considering whether this admonition now belongs under `The Earned Name`, since the coupling it warns about is a property of that name specifically.

docs/apps/register-dot-domain.md
```diff
@@ -70,31 +82,66 @@ When you run `playground deploy` and reach the `domain` prompt, enter the name y
...
 From there, the CLI registers the name on chain. If you deploy with the phone signer, each on-chain step is a separate approval in the Polkadot App, in this order:
 
-1. **Reserve domain**: Submits a DotNS commitment for the name without revealing it in the clear.
+1. **Reserve domain**: Submits a dotNS commitment for the name without revealing it in the clear.
 2. **Finalize domain**: Claims the name for your account.
 3. **Link content**: Points the name's `contenthash` at your uploaded bundle's CID, so the name now resolves to your Product.
 
 !!! note "The ~60-second pause is expected"
-    Between reserve and finalize, the deploy pauses for about 60 seconds. This is DotNS's commit-reveal window: the commitment is submitted first, then the name is claimed a short time later, so a watcher cannot see your desired name and race to register it ahead of you. The deploy is not stuck.
+    Between reserve and finalize, the deploy pauses for about 60 seconds. This is dotNS's commit-reveal window: the commitment is submitted first, then the name is claimed a short time later, so a watcher cannot see your desired name and race to register it ahead of you. The deploy is not stuck.
```
re-gius:
```suggestion
    Between reserve and finalize, the deploy pauses for about 60 seconds. Most of that is the tooling waiting for the commitment to settle: the contract requires only that the commitment be six seconds old before the name is claimed. The two-step handshake is what stops a watcher seeing your desired name and racing to register it ahead of you. The deploy is not stuck.
```

One related fact, if it is useful anywhere on this page: `MAX_COMMITMENT_AGE` = 1 days, so an abandoned commitment expires after a day and the handshake has to start over.

docs/apps/register-dot-domain.md
```diff
@@ -70,31 +82,66 @@ When you run `playground deploy` and reach the `domain` prompt, enter the name y
...
+## Register with the dotNS CLI
+
+The [dotNS CLI](/reference/apps/infrastructure/dotns/cli/) registers a name on its own, with no bundle and no deploy involved. Reach for it when you want to hold a name before the Product is ready, script a batch of names, or debug a registration that failed:
+
+```bash
+npm i -g @parity/dotns-cli
+dotns register domain --help
+```
+
+Registration runs the same commit-reveal handshake described above, so expect it to take a few minutes whichever tool drives it. The CLI can also preview `PopRules` eligibility for a proposed name and account, showing the band and the deposit before you submit.
```
re-gius:
```suggestion
Registration runs the same commit-reveal handshake described above, so expect a comparable pause whichever tool drives it. The CLI can also preview `PopRules` eligibility for a proposed name and account, showing the band and the deposit before you submit
```

Following from the note at line 90: since the six-second window is a contract parameter and identical for every tool, the variable part is how long each CLI waits, which makes "a few minutes" hard to source. Deferring to the figure already on the page avoids stating a second number that needs its own verification.

re-gius (commented):
I believe we should enforce a `DotNS` -> `dotNS` normalization and use dotNS everywhere

re-gius (commented):
I have a few more comments - still working on it

docs/apps/register-dot-domain.md
```diff
@@ -70,31 +82,69 @@ When you run `playground deploy` and reach the `domain` prompt, enter the name y
...
 2. **Finalize domain**: Claims the name for your account.
 3. **Link content**: Points the name's `contenthash` at your uploaded bundle's CID, so the name now resolves to your Product.
 
 !!! note "The ~60-second pause is expected"
-    Between reserve and finalize, the deploy pauses for about 60 seconds. This is DotNS's commit-reveal window: the commitment is submitted first, then the name is claimed a short time later, so a watcher cannot see your desired name and race to register it ahead of you. The deploy is not stuck.
+    Between reserve and finalize, the deploy pauses for about 60 seconds. Most of that is the tooling waiting for the commitment to settle: the contract requires only that the commitment be six seconds old before the name is claimed. The two-step handshake is what stops a watcher seeing your desired name and racing to register it ahead of you. The deploy is not stuck.
+
+!!! note "An abandoned commitment expires after a day"
+    A commitment is valid for `MAX_COMMITMENT_AGE`, one day. If a deploy fails between reserve and finalize and you come back later than that, the commitment is no longer claimable and the handshake starts over from the beginning.
```
re-gius:
```suggestion
    A commitment is valid for one day. If a deploy fails between reserve and finalize and you come back later than that, the commitment is no longer claimable and the handshake starts over from the beginning.
```

Sorry, this is my fault, I gave you `MAX_COMMITMENT_AGE` in the previous comment but that name only exists in the deployment script. A developer who goes looking for it on the contract will not find it.

Since the point of the note is the one-day expiry rather than the identifier, dropping it reads better and cannot go stale if a deployment is configured differently:

re-gius (approved):
LGTM - just one final comment

Issue #1747, opened 2026-08-28
Build guides diverge on SDK setup: import-specifier claim, stale pin, four install patterns, five account setups


ReinhardHatko:
The Build guides each set up the SDK their own way. Same task on every page — install, get a connected account, get a signer — five different shapes, four install patterns, and one stale version pin. Symptoms of there being no shared setup.

## 1. Three guides contradict the hub pages on import specifiers

`product-sdk/index.md:24` (and `build/index.md:132`):

> The import specifiers differ between the two: the umbrella exposes subpaths like `@parity/product-sdk/cloud-storage`, while the standalone package is `@parity/product-sdk-cloud-storage`. Switching styles means updating your imports.

Three guides still say the opposite:

- `build/read-chain-state.md:42`
- `build/sign-and-submit.md:45`
- `build/persist-data-locally.md:40`

> All import paths shown in this guide work with both options.

Their own snippets import from standalone specifiers (`@parity/product-sdk-chain-client`, `@parity/product-sdk-local-storage`), so a reader who picks the umbrella option and trusts that sentence gets an unresolvable import. This was corrected on the two hub pages in #1740; these three were missed.

## 2. Stale version pin

`build/pub-sub-off-chain-data.md:38`:

```bash
npm install @parity/product-sdk@0.8.0 @parity/product-sdk-statement-store@0.4.0
```

Current: `@parity/product-sdk` **0.25.0**, `@parity/product-sdk-statement-store` **0.6.7**.

`tutorials/shared-todo-app.md:28` is the model to copy here — it separates "verified against" from "still applies through", so the pin stays useful as versions move:

> The code in this tutorial was verified end to end in Polkadot Desktop against `@parity/product-sdk` v0.11.0 and `@parity/product-sdk-statement-store` v0.4.4. Every SDK surface it uses … is unchanged through `@parity/product-sdk` v0.23.0 and `@parity/product-sdk-statement-store` v0.6.5, so the code still applies on the current line.

## 3. Four ways to present the same install choice

| Page | Pattern |
|---|---|
| `read-chain-state`, `sign-and-submit`, `persist-data-locally` | bullets with nested code blocks, umbrella vs individual |
| `store-data-on-chain` | umbrella only, one line, no choice offered |
| `pub-sub-off-chain-data` | version-pinned, no choice offered |
| `product-sdk/index.md` | prose describing both |

`read-chain-state` adds a third install line for descriptors outside the two options, which is correct but breaks the shape again.

[...]

PR #1755, opened 2026-09-18
Fix apps docs sdk accuracy

### LINKED ISSUE #1742: checkAuthorization snippets read fields off the Result wrapper — auth check always fails
## Summary

Two code snippets read fields directly off the `Result` wrapper returned by `checkAuthorization`, instead of unwrapping it. Because `Result` carries no passthrough fields, the authorization check reads `undefined` and always takes the failure branch.

This is pre-existing (not introduced by #1740).

## The actual API

From `@parity/product-sdk-cloud-storage@0.11.1` (`dist/index.d.ts`), both call forms return a `Result`:

```typescript
// line 392 — client method
checkAuthorization(address: string): Promise<Result<AuthorizationStatus, CloudStorageAuthorizationError>>;

// line 438 — standalone export
declare function checkAuthorization(api: CloudStorageApi, address: string): Promise<Result<AuthorizationStatus, CloudStorageAuthorizationError>>;
```

And from `@parity/result`:

```typescript
type Result<T, E> = { ok: true; value: T } | { ok: false; error: E };
```

So `authorized`, `remainingTransactions`, `remainingBytes`, and `expiration` live on `.value`, never on the returned object itself.

## Affected files

**1. `docs/.snippets/code/apps/build/store-data-on-chain/check-authorization.ts`**

```typescript
const auth = await client.checkAuthorization(account.address);

if (!auth.authorized) {
  console.log(`No authorization found for ${account.address}`);
} else {
  console.log(`Remaining transactions: ${auth.remainingTransactions}`);
  ...
}
```

`auth.authorized` is `undefined`, so `!auth.authorized` is always `true`. The snippet always reports "No authorization found" regardless of the account's real state, and the `else` branch printing the quota is unreachable.

**2. `docs/.snippets/code/apps/tutorials/shared-todo-app/snapshot.ts` (lines 36-37)**

```typescript
const status = await readOnly.checkAuthorization(address);
if (!status.authorized) {
  throw new Error(
    `Account ${address} has no Bulletin Chain storage authorization — ...`,
  );
}
```

Same shape, but inside a `throw`, so this path fails unconditionally. Note the tutorial states its code is "confirmed working with Polkadot Desktop, `@parity/product-sdk` v0.11.0".

**3. `docs/apps/build/store-data-on-chain.md` (lines 112-115)** — the prose reinforces the wrong shape:

> The returned `AuthorizationStatus`:
> - **`authorized`**: `true` when an authorization record exists for the account.
> - **`remainingTransactions`**: Number of `store` calls remaining in the quota.

The returned value is a `Result` wrapping an `AuthorizationStatus`, not an `AuthorizationStatus`.

## Suggested fix

```typescript
const auth = await client.checkAuthorization(account.address);

if (!auth.ok) {
  console.error(auth.error.message);
} else if (!auth.value.authorized) {
  console.log(`No authorization found for ${account.address}`);
} else {
  console.log(`Remaining transactions: ${auth.value.remainingTransactions}`);
  console.log(`Remaining bytes:        ${auth.value.remainingBytes}`);
  console.log(`Expires at block:       ${auth.value.expiration}`);
}
```

and reword the prose in `store-data-on-chain.md` to say it resolves with a `Result` whose `.value` is an `AuthorizationStatus`.

## Note

`docs/apps/product-sdk/cloud-storage.md` already documents this correctly (checks `.ok`, then reads `.value.authorized`) — it can serve as the reference shape.

Worth a look at whether `getBulletinAllowanceStatus` is a better fit for these snippets: it returns `BulletinAllowanceStatus extends AuthorizationStatus` with derived `remainingBlocks` and `usable` fields, which is what both call sites are really trying to compute.

### LINKED ISSUE #1743: read-chain-state.md: createChainClient takes no RPC endpoints, and the page contradicts itself
## Summary

`docs/apps/build/read-chain-state.md` tells readers to pass RPC endpoints to `createChainClient`. That parameter does not exist — `ChainClientConfig` accepts `chains` and nothing else. The page also contradicts itself: line 78 says you supply endpoints, line 99 says you don't.

Pre-existing on `master`; not introduced by #1740.

## The bug

**Line 78:**

> Use `createChainClient` to supply your own descriptors **and RPC endpoints**, importing the chain descriptor objects you need directly instead of relying on a preset.

There is no way to do this. From `@parity/product-sdk-chain-client`:

```typescript
export interface ChainClientConfig<
    TChains extends Record<string, ChainDefinition> = Record<string, ChainDefinition>,
> {
    /** Named chain descriptors (PAPI `ChainDefinition` objects). */
    chains: TChains;
}
```

`createChainClient(config)` takes one parameter with no overloads, and `config.chains` is the only config field read anywhere in the implementation. The runtime path never involves an endpoint:

```javascript
const descriptor = config.chains[name];
const genesis = descriptor.genesis;
if (!genesis) throw new Error(`Descriptor for chain "${name}" has no genesis hash.`);
// → getHostProvider(genesis)
```

Descriptor → genesis hash → host provider. The host resolves the transport.

**Line 99 already states this correctly**, 21 lines later and about the same function:

> Connections are routed through the host at runtime, so you don't supply RPC endpoints yourself.

### Verified against

| Source | `ChainClientConfig` | config fields read |
|---|---|---|
| npm `latest` = 0.12.1 | `{ chains }` | `config.chains` |
| npm `dev` = 0.0.0-dev.312.0 | `{ chains }` | `config.chains` |
| `paritytech/product-sdk` @ `main` | `{ chains }` | `config.chains` |

No `wsUrl`, `endpoint`, `getWsProvider`, or any RPC identifier appears in `chain-client/src/clients.ts` at HEAD. The `@parity/product-sdk-descriptors` package does ship `wsUrl` values, but only inside `chains/*/.papi/polkadot-api.json` — PAPI **codegen** config used at build time to fetch metadata. The runtime descriptor export (`chains/*/generated/dist/index.js`) contains `genesis` and no URL.

## Suggested fix

Line 78 — drop the phrase:

> Use `createChainClient` to supply your own descriptors, importing the chain descriptor objects you need directly instead of relying on a preset.

**Lines 54 and 58** (lower priority, wording only). These say the preset "comes preconfigured with descriptors and RPC endpoints":

> The Preset path (`getChainAPI`) ... comes preconfigured with descriptors and RPC endpoints for supported environments.

> `getChainAPI(env)` returns a client preconfigured with the descriptors and RPC endpoints for the requested environment.

Read as "you don't need to supply these," the takeaway is correct, so these are not misleading about what to *do*. But they are inaccurate about what the preset contains — it holds descriptor imports only:

```javascript
paseo: () => Promise.all([
  import('@parity/product-sdk-descriptors/paseo-asset-hub'),
  import('@parity/product-sdk-descriptors/paseo-bulletin'),
  import('@parity/product-sdk-descriptors/paseo-individuality')
]),
```

They also make line 78 *believable*: a reader told endpoints are part of the model has no reason to doubt the BYOD sentence. Worth tightening to "preconfigured with the descriptors for the requested environment; the host resolves the connection from the descriptor's genesis hash."

## Upstream: the same error is in the SDK's JSDoc

The docs faithfully copied the SDK. Both phrases are still present in `paritytech/product-sdk` at `main`:

```
product-sdk/packages/chain-client/src/clients.ts:52
 * Create a multi-chain client with user-provided descriptors and RPC endpoints.

product-sdk/packages/chain-client/src/index.ts:7
 * `getChainAPI` is the zero-config path with built-in descriptors and RPC endpoints
```

`clients.ts:52` is word-for-word line 78 here; `index.ts:7` is near word-for-word line 54. Both contradict the `ChainClientConfig` doc comment in `types.ts` in the same package, which says the opposite:

> Provide named chain descriptors. The SDK routes all connections through the host provider, **so no RPC endpoints are required**, and only chains the host routes are reachable.

Worth fixing upstream as well, or the phrasing gets copied into docs again on the next pass. Happy for someone with context on that repo to raise it there.

## Not verified

Whether an unmerged branch or PR in `product-sdk` adds an endpoint field, and whether `@parity/product-sdk-host` accepts endpoint configuration by some other route. Neither would make line 78 correct as written, since it is specifically about what you pass to `createChainClient`.

### LINKED ISSUE #1744: setup-app.ts: createApp name yields zero accounts, and wallet.connect() is unguarded
## Summary

`docs/.snippets/code/apps/build/store-data-on-chain/setup-app.ts` is four lines and has two problems. It is the setup snippet every later snippet on [Store Data on Chain](https://docs.polkadot.com/apps/build/store-data-on-chain/) imports, so both propagate through the page.

Pre-existing; the snippet is not touched by #1740.

```typescript
import { createApp } from '@parity/product-sdk';

export const app = await createApp({ name: 'my-product' });
await app.wallet.connect();
```

## 1. `name: 'my-product'` yields zero accounts

`createApp` passes `name` through as the signer's `dappName`:

```typescript
// packages/sdk/src/core/createApp.ts
const localKvStore = await createLocalKvStore({ prefix: config.name });
...
    dappName: config.name,
```

Per the umbrella CHANGELOG, that value is the dotNS identifier the host derives the account from:

> The `dappName` you pass to `createApp({ name })` or `new SignerManager({ dappName })` is now also the dotNS identifier the host derives the product account from. `.dot` is appended automatically if missing. If your `dappName` isn't a valid registered dotNS identifier, the host will reject the derivation and `connect()` will resolve with `[]`.

`my-product` is a bare label, so the host resolves `my-product.dot`. Unless the reader owns that name, derivation fails and `connect()` succeeds with zero accounts.

## 2. `wallet.connect()` is unguarded and its result is discarded

`app.wallet.connect()` throws rather than returning a `Result`:

```typescript
// packages/sdk/src/core/createApp.ts — createWalletApi
async connect(): Promise<{ accounts: Account[] }> {
    const result = await signerManager.connect();
    if (!result.ok) {
        throw new Error(result.error.message);
    }
```

So outside a Host this line throws. Inside a Host with an unregistered name it returns `{ accounts: [] }` — and because the snippet discards the return value entirely, there is nothing to check and no way to notice.

## Downstream effect

`hello-bulletin.ts`, `retrieve-data.ts`, and the rest import `app` from here. `CloudStorageApi.upload` is documented as requiring "a wallet to be connected and an account selected", so with zero accounts the whole page's flow fails at the first upload, with the cause three sections earlier.

## Suggested fix

Guard the connect, check for the empty list, and say what `name` has to be:

```typescript
import { createApp } from '@parity/product-sdk';

// `name` is also your dotNS identifier: the Host derives the user's product
// account from it, appending `.dot` if missing. Use a name you have registered,
// or connect() will succeed with zero accounts.
export const app = await createApp({ name: 'my-product.dot' });

const { accounts } = await app.wallet.connect();
if (accounts.length === 0) {
  throw new Error('Host derived no account for this name — is it a registered .dot name?');
}
```

## Related

- #1742 — other snippets on this page that read fields off a `Result` wrapper.
- #1740 added exactly this warning to `docs/apps/product-sdk/index.md` ("`name` is also your dotNS identifier") and guarded the equivalent snippet there. This snippet is now inconsistent with that page.

### LINKED ISSUE #1745: Storage options callout has no canonical home; reference pages cite it at two URLs
## Summary

The "Storage options for your Product" callout has no canonical home, so reference pages cite it at two different URLs.

The callout itself is fine: it lives in `docs/.snippets/text/apps/storage-options.md` and renders on three build guides — `store-data-on-chain.md`, `pub-sub-off-chain-data.md`, `persist-data-locally.md`, all at line 15. Single source, no drift. (#1740 extracted it; before that there were three inline copies and one had already drifted, mislabelling "Local KvStore (this page)".)

## The problem

Three reference pages link to the callout **by name**, pointing at two different build guides:

| File | Links to |
|---|---|
| `reference/apps/protocol/truapi/statement-store.md:13` | `/apps/build/pub-sub-off-chain-data/` |
| `reference/apps/infrastructure/statement-store/index.md:15` | `/apps/build/pub-sub-off-chain-data/` |
| `reference/apps/infrastructure/bulletin-chain/index.md:38` | `/apps/build/store-data-on-chain/` |

No link is broken — each target renders the box. But "Storage options for your Product" resolves to a different page depending on which reference page you arrive from, and each of those pages is a task guide about one storage layer, not a comparison of all of them. A reader sent from the Bulletin Chain reference to "storage options" lands on the Bulletin how-to.

## Suggested fix

`apps/concepts/data-placement.md` ("Where to Store Data") already covers this exact decision across the same four layers, with a comparison table and a section per option. It is the natural destination for a cross-reference.

- Point the three reference-page links at `/apps/concepts/data-placement/`.
- Keep the build-guide callout as an inline shortcut for readers already on a task page.

That leaves one URL to cite and one page to maintain when the storage layers change.

Low priority — coherence, not a defect.

### LINKED ISSUE #1746: persist-data-locally: clear() is a no-op, and the app.localStorage / createLocalKvStore examples are not composable
Two problems on [Persist Data Locally](https://docs.polkadot.com/apps/build/persist-data-locally/). Both pre-existing.

## 1. `app.localStorage.clear()` does nothing

`persist-data-locally.md:114`:

> `app.localStorage.clear()` removes every key scoped to your Product. It is equivalent to calling `remove()` on each key individually.

It is a no-op. From `@parity/product-sdk@0.25.0`, [`createApp.ts#L134-L137`](https://github.com/paritytech/product-sdk/blob/713ae3378b4aeddbbb3843454523e0465f7f5933/product-sdk/packages/sdk/src/core/createApp.ts#L134-L137):

```typescript
clear: async () => {
    // LocalKvStore doesn't have clear - this is a no-op
    log.debug("clear() is not supported in container storage mode");
},
```

It also could not work as described: `HostLocalStorage` has no enumeration method (no `keys`, `list`, `entries`), so there is nothing to iterate over. Note `HostLocalStorage.clear(key: string)` takes a key — at that layer "clear" means remove one key, not clear all.

`remove-and-clear.tsx` ships a button for it:

```tsx
async function handleClear() {
  // Removes every key scoped to this Product.
  await app.localStorage.clear();
}
```

It resolves successfully, so nothing surfaces the failure. A reader wires this up and ships it.

`persist-data-locally.md:119` has the same implication: "`app.localStorage.clear()` and `store.remove()` are scoped to your Product."

**Fix:** drop the claim and the button, or document `clear()` as unimplemented. Filing upstream separately.

## 2. The examples are not composable

The page mixes two objects with different key spaces, without saying so:

- `createLocalKvStore({ prefix: 'feature' })` → `<productId>:feature:<key>`
- `app.localStorage` → `createApp` builds it as `createLocalKvStore({ prefix: config.name })`, so `<productId>:<name>:<key>`

Each snippet is correct alone. Written with the prefixed store and read back via `app.localStorage`, you get `null` — no error, just a miss.

Line 89 explains prefix stacking correctly, but only for the direct-store case. Nothing says `app.localStorage` is itself a prefixed store, or that the two cannot see each other's keys.

**Fix:** state that `app.localStorage` is a store prefixed by `createApp`'s `name`, and that separately created stores are separate key spaces.

### LINKED ISSUE #1748: reference/apps/index.md: "Product SDK" links to the Polkadot SDK glossary entry
`docs/reference/apps/index.md:20` links "Product SDK" to the **Polkadot SDK** glossary entry:

```markdown
the [Product SDK](/reference/glossary/#polkadot-sdk) that exposes typed methods
```

These are different things. Polkadot SDK is the Rust framework for building chains and parachains; Product SDK is the TypeScript SDK for building Products. The sentence is specifically about the layered stack a Product sits on top of, so the link sends the reader to the wrong layer of the very stack being explained.

## Fix

The glossary has `## Polkadot SDK` and `## Product`, but no Product SDK entry, so there is no correct anchor to swap in. The natural target is the section overview added in #1740:

```markdown
the [Product SDK](/apps/product-sdk/) that exposes typed methods
```

Alternatively, add a `## Product SDK` glossary entry and point at that — worth considering given the two names are easy to confuse and the glossary already carries `Polkadot SDK`.

## Note

The bad link predates #1740; `docs/reference/apps/index.md` is not touched by it. It is newly *fixable*, though, because that PR created `/apps/product-sdk/` as a real destination.

bucanero:
## Summary

Fixes the Apps documentation issues reported by @ReinhardHatko. Every claim was verified against the published `@parity/product-sdk*` type definitions and source, not against the SDK's JSDoc — which is wrong in the same way the docs were, and is where some of this originally came from.

All seven reports were valid.

## What changed

| Issue | Fix |
|---|---|
| #1742 | `checkAuthorization` returns a `Result`. The snippet read `auth.authorized` off the wrapper, so it always reported "No authorization found" and the quota branch was unreachable. |
| #1743 | `createChainClient` takes no RPC endpoints — `ChainClientConfig` is `{ chains }`. The page also contradicted itself 21 lines apart. |
| #1744 | `createApp`'s `name` is the dotNS identifier the Host derives the account from. An unregistered name resolves `connect()` with zero accounts instead of failing; the snippet discarded the result, so every upload on the page failed with the cause three sections earlier. Guard added, and the guides aligned with `product-sdk/index.md`. |
| #1745 | Three reference pages cited "Storage options for your Product" at two URLs, neither a comparison page. All now point at Where to Store Data. |
| #1746 | `app.localStorage.clear()` does not exist — upstream removed it after it shipped as a no-op. Documented method and its React button are gone. Also warns that `app.localStorage` and prefixed stores are separate key spaces. |
| #1747 (items 1–2) | Three guides claimed all import paths work with either install option; their snippets use standalone specifiers, and two do not match their leaf package. `pub-sub` pinned versions several breaking releases behind. |
| #1748 | "Product SDK" linked to the Polkadot SDK glossary entry. |

## Also fixed (not reported)

`publish-statement.ts` branched on `if (accepted)` where `publish` returns a `Result`, so it reported every publish as accepted, including rejections — the exact trap the shared-todo-app tutorial warns about. `channel-presence.ts` called `onChange`/`readAll` keys `name` when they are hex hashes.

## Not in scope

- #1747 items 3–4 (four install patterns, five account setups) are a restructure rather than a factual fix, and are left for a follow-up. The factual inconsistencies underneath are fixed here.
- `sign-and-submit.md:27` still carries a stale "confirmed working with v0.9.0" note. That is a verification claim and these snippets have not been re-run, so it was left alone.
- Upstream: `product-sdk`'s `clients.ts:52` and `index.ts:7` still carry the wrong "and RPC endpoints" phrasing, contradicting `types.ts` in the same package. Without an upstream fix it gets copied back in on the next pass.


docs/apps/build/read-chain-state.md
```diff
@@ -51,11 +51,11 @@ The descriptors package exposes typed `ChainDefinition` objects through subpath
 
 ## Connect to a Chain
 
-The SDK provides two connection paths. The Preset path (`getChainAPI`) is the fastest way to get a working client — it comes preconfigured with descriptors and RPC endpoints for supported environments. Use the BYOD path (`createChainClient`) when you need explicit control over which chains and descriptors your Product uses.
+The SDK provides two connection paths. The Preset path (`getChainAPI`) is the fastest way to get a working client — it comes preconfigured with the descriptors for supported environments. Use the BYOD path (`createChainClient`) when you need explicit control over which chains and descriptors your Product uses. Neither path takes RPC endpoints: the Host resolves the connection from each descriptor's genesis hash.
 
 ### Connect Using a Preset
 
-`getChainAPI(env)` returns a client preconfigured with the descriptors and RPC endpoints for the requested environment.
+`getChainAPI(env)` returns a client preconfigured with the descriptors for the requested environment.
```
ReinhardHatko: The RPC-endpoint half of #1743 is fixed here and in the two other spots. Two items from that issue are still open, so **#1743 should not be closed when this merges.**

**This line — `env` still reads as a request.**

> `getChainAPI(env)` returns a client preconfigured with the descriptors for the requested environment.

`env` is optional, and host discovery runs first and unconditionally:

```typescript
export async function getChainAPI(): Promise<ChainClient<PresetChains<"paseo">>>;
export async function getChainAPI<E extends Environment>(env: E): Promise<...>;

export async function getChainAPI(envArg?: Environment): Promise<any> {
    const discovery = await getHostChainInfo(Object.values(CHAIN_IDENTIFIERS));
    const env = await resolveEnvironment(envArg, discovery);
```

When `env` is passed it is cross-checked rather than obeyed, and disagreeing with the host is fatal: `if (matched && matched !== env) throw new EnvironmentMismatchError(env, discovery.network)`. A comment in `resolveEnvironment` states the intent — "Hosts mint their own network ids, so matching is by genesis, never by network string."

So `env` is an assertion the host can veto, which is what `product-sdk/index.md` already tells readers ("the Host … selects the network"). The no-arg `getChainAPI()` overload — the call that actually expresses that — is not documented on this page.

**Line 74 — two of the six listed values do not exist.** Outside the diff, so quoting rather than anchoring:

> `getChainAPI` accepts the `Environment` values `polkadot`, `kusama`, `paseo`, `devnet`, `local`, and `westend`. Only `paseo` and `devnet` are wired up at the moment; calling an environment that is not yet live throws at runtime.

`chain-names.ts` defines the union and neither `local` nor `westend` is a member — passing either is a compile error, not the runtime throw the second sentence describes. The availability claim in that same sentence, and the matching Limitations bullet at line 206, are worth re-checking against `AVAILABLE_ENVIRONMENTS` in `presets.ts` while you are in there.

Of the two, line 74 is the one a reader hits fastest: it is a concrete list they will copy from.

ReinhardHatko (approved):
Verified every claim against the published packages rather than the JSDoc — umbrella 0.29.0, statement-store 0.6.11, chain-client 0.12.5, signer 0.14.6, local-storage 0.3.11. All of it holds, including the parts that were easy to get wrong:

- The root re-export split is exactly right: `createChainClient` and `SignerManager` are at the umbrella root, `getChainAPI` and `destroyAll` are not.
- `LocalStorageApi` is precisely the five methods the new note lists. Deleting the button rather than rewording was the right call — `clear()` is a type error now, not a no-op.
- Version pins match npm exactly.
- `estimateAuthorization` returns a plain object, so leaving that snippet tail unchanged is correct.

The `channel-presence.ts` fix deserves particular credit. The `.d.ts` names the parameter `channelName` and the SDK's own JSDoc example calls it `name` — both misleading. The implementation disagrees:

```javascript
const hashKey = topicToHex(createChannel(channelName));
this.nameToHash.set(channelName, hashKey);
this.updateChannel(hashKey, timestamped);
```

and the source comment confirms it: "Keys are hex-encoded channel hashes. Use `read` for lookup by human-readable name." Reaching for `channels.read(name)` is the correct resolution.

The two self-reported finds are both real — `if (accepted)` on a `Result` is always truthy, so every rejection was reported as accepted.

One non-blocking note below on #1743, which is half fixed. Approving regardless; the remaining items are one-liners and can land here or separately.

Also worth flagging outside this PR: the deferrals are the right calls and are stated rather than quietly dropped. Leaving `sign-and-submit.md:27`'s stale "confirmed working" note alone because the snippets have not been re-run is good discipline.

